PR #1, opened 2019-11-10
Get rid of scaling log message

GreenAsJade:
It's quite noisy in the console, gets in the way of other debug.


anoek:
Indeed, thanks!

PR #2, opened 2020-03-25
Show a pass as a '.' in the "move number" in the move tree

GreenAsJade:
As a moderator, I need this all the time, so that when I get called to "timewasting" reports I can see at a glance where the reporter first passed, instead of trawling through the game to see where timewasting started.

I'd thought to dim the stone which is a pass, but that's a much deeper change (requiring modification to placeBlackStone, placeWhiteStone, so I went with this lightweight version.

Screen Shot 2020-03-25 at 2 02 26 pm (2)


anoek:
Nice, thanks!

PR #4, opened 2020-06-06
Add support for iOS 10 in minimized scripts

guest951:


anoek:
Thanks for the fix in both places!

guest951:
It seems that you changed the version number just a little too soon (because your correction was on the server side scripts, but the problem is rather on the client side scripts).

PR #6, opened 2020-06-21
External goban image 

Adam314159:
**WARNING: This PR is co-dependant with [this one](https://github.com/online-go/online-go.com/pull/1179) on online-go.com**

---

The basic idea is to allow users to style their goban with a custom image in the form of supplying a link to an external picture. 

- I have changed the look of the plain board preview in the selector, to more intuitively represent the cutomizable nature
Screenshot_2020-06-21 OGS

- Upon selecting the custom option an input form appears together with the color selectors, once the text is replaced with an image link, it will automatically update the board background. 
Screenshot_2020-06-21 OGS(1)

- in the future I was thinking of creating a similar option for the stones, together with two buttons allowing for playing one-color-go and blind-go easily. Let me know what you think about that.

---

Please note, this is my first barbecue with React. I mostly wanted to have a look around with a simple/fun project. If the code is not up to par, or if you think it was a stupid idea to begin with, don't be afraid to tell me to shove it :)

### DIFF
--- a/src/GoTheme.ts
+++ b/src/GoTheme.ts
@@ -19,11 +19,13 @@
 export interface GoThemeBackgroundCSS {
     'background-color'?: string;
     'background-image'?: string;
+	'background-size'?: string;
 }
 
 export interface GoThemeBackgroundReactStyles {
     'backgroundColor'?: string;
     'backgroundImage'?: string;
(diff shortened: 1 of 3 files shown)


anoek:
Fun, thanks!

PR #7, opened 2020-07-12
Don't issue error messages on "stone_already_placed_here" exception

GreenAsJade:
Sentry keeps getting "stone_already_placed_here" reports, and they appear in the console.

I don't think there's any reason for this: the goban raises this exception to bomb out of onClick processing, but it's not really an error.

this.errorHandler() already has a check to ignore this exception, but unfortunately at the level that this.errorHandler() is called there is another place (console.info(e) that is causing it to be reported.  Hence the need for this, I think.


anoek:
Sorry for the delay on merging this, thanks for the fix!

PR #12, opened 2020-11-11
Added start-time and end-time to GoEngineConfig declaration

GreenAsJade:
start-time and end-time are in the GoEngineConfig we get from the server, so declaration needed


anoek:
Thanks!

PR #14, opened 2020-11-21
Return to previous mode and board position when exiting score estimator

JDNdeveloper:
Previously when pressing "Back to Game" (or `esc`) in score estimator mode you would be taken back to the last official move and placed in the `play` mode. This change returns you to the previous board mode and position.


anoek:
Thanks!

PR #15, opened 2020-11-29
Fix stop estimating score

JDNdeveloper:
Fixes these issues (will need to manually close these since they're in the `online-go` repo):
https://github.com/online-go/online-go.com/issues/1290
https://github.com/online-go/online-go.com/issues/1291

This fixes the bugs by doing two things:
1. Return to `play` mode and jump to official move unless we were previously in `analyze` or `conditional` mode (i.e. only allow this "feature" where we think it's safe, rather than allowing for exposing latent bugs when returning to different modes which was the cause of the above issues). This should fix the "staged submit move" by just jumping back to the official move.
2. Allow conditional mode even if it's the current players turn if we were previously in estimate score mode.

### DIFF
--- a/src/GobanCore.ts
+++ b/src/GobanCore.ts
@@ -3038,9 +3038,12 @@ export abstract class GobanCore extends TypedEventEmitter<Events> {
             this.redraw(true);
             this.emit("update");
         } else {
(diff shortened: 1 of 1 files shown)


anoek:
Awesome, thanks for the fix!

PR #18, opened 2020-12-09
Fixed Goban error message on mobile and puzzle ko bug

JDNdeveloper:
Fixes the following bugs:
- Ko's were not respected in puzzles if they occured within the first 3 moves. This is because pre-placed stones in the puzzle do not count towards the move count, which means Ko's can occur within the first 3 moves of the puzzle which is impossible in real games/reviews. Fixed this by ignoring the `this.cur_move.move_number > 2` check if we are in puzzle mode.
- Fixed Goban error message disappearing as soon as it appears on mobile due to "phantom click". Prevented this by delaying the message from rendering by 100ms, allowing the "phantom click" to occur before the `click` event handler has been registered for the message.

This screen capture shows both bugs that were fixed in this PR: https://streamable.com/x3qwqf

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -1000,7 +1000,9 @@ export class GoEngine {
                     }
                 }
 
-                if (checkForKo && !this.allow_ko && this.cur_move.move_number > 2) {
+                if (checkForKo && !this.allow_ko &&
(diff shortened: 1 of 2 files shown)


anoek:
Thanks for the fixes!

PR #19, opened 2020-12-11
Added variation stone transparency support to Goban

JDNdeveloper:
This change allows for a custom variation stone transparency level. It will have no effect until the corresponding OGS PR merges: https://github.com/online-go/online-go.com/pull/1317


anoek:
Cool, thanks!

PR #20, opened 2020-12-12
Respect config.mode in constructor

benjaminpjones:
After a long time trying to figure out why I couldn't place stones in the test, I finally realized that `GobanCore`'s constructor didn't do anything with `config.mode`.  There was also inconsistent typing that I decided to take care of regarding `GobanCore.mode` and `GobanConfig.mode`.  Once I tightened up the types, it was revealed that `GobanCore.has_new_official_move` isn't used, so I removed it.

This change should not affect online-go.com as it is using Goban.setMode() in most places to set the mode, but I was not able to verify this.  Is there a way to run dev.beta.online-go.com using a dev version of Goban?

### DIFF
--- a/src/GobanCore.ts
+++ b/src/GobanCore.ts
@@ -56,7 +56,7 @@ export const MARK_TYPES:Array<keyof MarkInterface> = ["letter", "circle", "squar
 
 let last_goban_id = 0;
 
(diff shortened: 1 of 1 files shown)


anoek:
Cool, thanks!

PR #23, opened 2020-12-28
Show '?' on latest move if there is an undo request for it.

GreenAsJade:
Some people would like [a visual indication if they have received an undo request](https://forums.online-go.com/t/wish-better-noticeable-undo-requests/964/70?u=eugene).

This change provides a preference setting to indicate that this is desired, and puts a '?' on the latest move if there is an undo request on it.

It does _not_ try to colour the '?' red as some requested, because that is an absolute can of worms - the 'standard' black stone has a white highlight on it, which causes a red question mark to look bad-contrast.    Not to mention what happens with themes.

This implementation assumes that the colour that has been selected for other marks to be drawn on the stones is the right one.

### DIFF
--- a/src/GobanCanvas.ts
+++ b/src/GobanCanvas.ts
@@ -1644,16 +1644,30 @@ export class GobanCanvas extends GobanCore  {
                         draw_last_move = false;
                     }
                     else {
-                        ctx.beginPath();
-                        ctx.strokeStyle = color;
-                        ctx.lineWidth = this.square_size * 0.075;
-                        let r = this.square_size * 0.25;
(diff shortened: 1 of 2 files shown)


GreenAsJade:
What it looks like:

GreenAsJade:
An example of the problem of trying to colour the question mark:

Screen Shot 2020-12-28 at 12 55 24 pm (2)

anoek:
Very cool, nice job, thanks!

Spoekenkieker:
You’re right, B/W is MUCH better. Thanks for doing this!

PR #24, opened 2020-12-28
Fix semicolons and unnecessary assignment to undefined - make tslint happy

GreenAsJade:
Note: I checked that not-assigning to undefined does result in the variable being `=== undefined`, just as if it were assigned.


anoek:
Thanks for the cleanup!

PR #25, opened 2021-01-05
Fix sgf dead stones

### LINKED ISSUE #27: Uploaded SGF territory marking is incorrect
There are two issues:

1. For games that end in resignation, no territory should be marked, however this is not the case (e.g. https://online-go.com/game/29813935):
image

2. For games that do go to scoring, any territory with dead stones is not marked (e.g. https://online-go.com/game/29322975):
image

Right now it looks like the client just uses a basic heuristic to mark any unambiguous territory, but leaves the rest alone, which is not ideal.

I'm not sure if SGFs include any territory marking data (i.e. if the players manually marked dead stones), but it does not appear that the server provided game data includes this. If SGFs do include this, it would likely require a server side change to pass this information on to the client.

I think the best that can be done with a client-only change is not marking territory when a game ends in resignation.

EDIT: ~~Looks like KGS SGFs (which is where these were downloaded from) do not include dead stone marking despite the SGF format supporting it~~. So I think the real solution here is for a "smarter" client-side dead-stone marking algorithm, or integrating this with the upcoming "AI scoring" feature.

EDIT 2: Nevermind, KGS SGFs do include the `TW` and `TB` properties. So it would be cool if OGS respected those when displaying SGFs (and also set them when exporting SGFs from OGS!). Since this requires server side changes, for now I think the easiest solution is to try to use a smarter client-side marking algorithm (as stated previously), probably just using the basic score-estimator is good enough (or at least better than the current solution)?

JDNdeveloper:
Fixes #27 

This change uses the score estimator to find dead stones in SGF uploaded games, since they don't include `gamedata.removed` (even if the SGF did include dead stones). This makes the final territory marking more accurate.

In the resignation case, there was just a small bug where `outcome: resign` instead of `outcome: Resignation` in the game data, causing it to improperly render marks on the board for SGF games that resulted in resignation.

Both of these changes should now make SGF final board state rendering much closer to their native OGS game counterparts.

Before (when game ends in resignation):
image

After (when game ends in resignation):
image

Before (when game ends with score):
image

After (when game ends with score):
image

Once OGS server-side supports parsing dead stones from SGFs the `gamedata.removed` string will be set in the game payload, and then this logic won't trigger and we'll properly use that instead. (EDIT: this statement may not be out of date since I added TW/TB support to `parseSGF`).

EDIT: I added `TW`/`TB` parsing support, so if those properties are present we'll defer to that rather than the score estimator.

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -29,6 +29,9 @@ import { AdHocPackedMove } from './AdHocFormat';
 import {_} from "./translate";
 
 
+declare const CLIENT:boolean;
+declare const SERVER:boolean;
+
 export const AUTOSCORE_TRIALS = 1000;
 export const AUTOSCORE_TOLERANCE = 0.30;
 
@@ -500,7 +503,8 @@ export class GoEngine {
         if (config.removed) {
             removed = this.decodeMoves(config.removed);
         }
-        if (!this.territory_included_in_sgf &&
+        if (CLIENT &&
+            !this.territory_included_in_sgf &&
             typeof(config.removed) === "undefined" &&
             config.original_sgf &&
             (/[0-9.]+/.test(self.outcome))) {
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -499,8 +499,10 @@ export class GoEngine {
         if (config.removed) {
             removed = this.decodeMoves(config.removed);
         }
-        if (typeof(config.removed) === "undefined" && config.original_sgf &&
-            config.outcome !== "Resignation") {
+        if (typeof(config.removed) === "undefined" &&
+            config.original_sgf &&
+            typeof(config.outcome) !== "undefined" &&
+            (/[0-9.]+/.test(config.outcome))) {
             // 2021-01-05: Game data for SGF uploaded games currently don't include
             // removed stones, so we use score estimator to find probably dead stones to get a
             // closer approximation of what territories should be marked in the final board
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -506,8 +506,9 @@ export class GoEngine {
             // closer approximation of what territories should be marked in the final board
             // position.
             //
-            // NOTE: Some Go clients do include dead stones in their SGFs, so this is something
-            // we may be able to support in the future.
+            // NOTE: Some Go clients do include dead stones in their SGFs, however we ignore them,
+            // so this is something we may be able to support in the future with a server side
+            // change so we wouldn't require this hack.
             let se = new ScoreEstimator(this.goban_callback, this, AUTOSCORE_TRIALS, AUTOSCORE_TOLERANCE);
             removed = this.decodeMoves(se.getProbablyDead());
         }
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -29,6 +29,9 @@ import { AdHocPackedMove } from './AdHocFormat';
 import {_} from "./translate";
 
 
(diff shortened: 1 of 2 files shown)


anoek:
Fixes #27

src/GoEngine.ts
```diff
@@ -492,8 +496,24 @@ export class GoEngine {
...
+            typeof(config.removed) === "undefined" &&
+            config.original_sgf &&
+            (/[0-9.]+/.test(self.outcome))) {
+            // Game data for SGF uploaded games don't always include removed stones, so we use score
+            // estimator to find probably dead stones to get a closer approximation of what
+            // territories should be marked in the final board position.
+            //
+            // NOTE: Some Go clients (not OGS, at least for now) do include dead stones in their
+            // SGFs, which we respect if present and skip using the score estimator.
+            let se = new ScoreEstimator(this.goban_callback, this, AUTOSCORE_TRIALS, AUTOSCORE_TOLERANCE);
```
anoek, line 513: This line causes the server to kick off the ScoreEstimator on the server side every time an SGF is loaded without scoring, so we'd either need to check if we're in operating on the server and not run this, or simply remove this. I'm slightly on the side of not run this at all, for a couple of reasons: 1) if the sgf didn't set the territory, not running this is the "pure" way to see that SGF - and if the user wants the estimation they can always click the estimator button. 2) I'm hoping to ditch the score estimator all together in favor of sending scoring requests off to the server and using KataGo to do a presumably *much* better job than the score estimator, so wiring that up might be out of the scope of the goban project.

anoek:
This is great @JDNdeveloper :)

I am seeing a parse error arise when trying to load games that report a score result but that do not have the territory fields defined anywhere though. 

[29322975-sgf-deje-tony11-without-scores.sgf.txt](https://github.com/online-go/goban/files/5808993/29322975-sgf-deje-tony11-without-scores.sgf.txt)

[28179211-012-Chrome-anoek.sgf.txt](https://github.com/online-go/goban/files/5809004/28179211-012-Chrome-anoek.sgf.txt)

JDNdeveloper:
@anoek I'm not seeing a parser error for either of those files. Is this only occurring on upload? Since SGF uploads don't work on the dev site I upload on the beta site and then load the games on the dev site.

anoek:
hmm could be, I'll look into it

JDNdeveloper:
@anoek How can I check if we're server-side or client-side in the code? Is there a canonical way to do that in goban/ogs code or should I just use some basic heuristic like this: https://stackoverflow.com/questions/13644303/how-to-check-if-js-code-is-running-on-node-server-or-on-the-client

I'd prefer to leave the score estimation in for now, as skipping this step won't actually cause "pure" display since it will still mark unambiguous territories and lead to nonsensical board markings (see screenshot 3 in description). Alternatively we'd need to plumb through `typeof(config.removed) === "undefined"` to where the final board state is rendered and get it to NOT mark board positions if that condition is true (similar to what we do for "resignation" or "timeout" results), which is messy.

I'm happy to go through and rip this out once the KataGo score estimator is implemented.

anoek:
That did the trick, thanks :) Yeah the error was coming from the score estimator not having been initialized yet on the server side, so that fixed the upload issue too.

Thanks for working through it, and thanks for the patch!

anoek:
For any future readers `CLIENT` and `SERVER` booleans are set globally at build time, thanks @JDNdeveloper for rediscovering that :)

JDNdeveloper:
@anoek can you push the v0.5.34 version bump commit?

anoek:
Done

PR #28, opened 2021-01-06
Parsed outcome from SGF on load if missing

### LINKED ISSUE #26: Uploaded SGF "winner" text is wrong if SGF contains comments before "RE" (result) tag
Instead of showing "White/black wins by resignation" it only shows "White/black wins by". For example in this game: https://online-go.com/game/29322976

image

JDNdeveloper:
Fixes #26

Before:
image

After:
image

Another benefit that you can see from the screenshot is that we properly recognize that territory shouldn't be scored since it's a "timeout".

Opened new issue to track the server side change still required to fix the SGF library page showing "?" (which actually looks like this should fix as well, so hopefully a noop after deploy): https://github.com/online-go/online-go.com/issues/1348

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -2115,6 +2115,45 @@ export class GoEngine {
                                 if (val[0].toLowerCase() === "w") {
                                     self.winner = "white";
                                 }
+
+                                if (self.outcome === "") {
+                                    let result;
+                                    let match = val.match(/[BW]\+(.*)/);
+                                    if (match === null) {
+                                        result = val;
+                                    } else {
+                                        result = match[1];
+                                    }
+
+                                    if (match !== null && !isNaN(result as any)) {
+                                        // There's a numeric score.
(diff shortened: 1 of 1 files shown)


anoek:
Can you link or attach an SGF you were having trouble with? This might be a good fallback, but it seems like the root cause of the issue is in our parser perhaps?

anoek:
Never mind you included links on the issue, I was lost.

anoek:
I *think* this code would be better moved to here:

https://github.com/online-go/goban/blob/master/src/GoEngine.ts#L2109-L2120

JDNdeveloper:
@anoek isn't that where this code already is?

anoek:
Apparently I need more coffee, yes you're absolutely right.

anoek:
This is great, thanks for the fix! I see the corresponding issue on the server side, I think what I'll do is switch over to using this code on the server side in time - though not right now since that's a bit of an undertaking.

PR #29, opened 2021-01-30
Fixed unrecognized resignation outcome string

JDNdeveloper:
Game data has `outcome: 'r'` for these SGF uploaded games which I've never seen before, leading to incorrectly marking territory:
https://online-go.com/game/30411889
https://online-go.com/game/28824220

The SGFs seem to be properly formatted, so not sure what's going wrong in the server side parser. @anoek Did something recently change? I've never seen `r` before as an outcome.

This fixes the issue by handling the new outcome and correcting it to a recognized outcome type so that territory isn't marked.

Before (1):
Screenshot 2021-01-30 085812

After (1):
image

Before (2):
Screenshot 2021-01-30 085850

After (2):
image

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -1515,8 +1515,8 @@ export class GoEngine {
             delete config["ladder"];
         }
 
-        if (config.outcome === "resign") {
-            // SGF games have this non-standard outcome, so we correct it.
+        if (config.outcome === "resign" || config.outcome === "r") {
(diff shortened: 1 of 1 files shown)


anoek:
Interesting, nothing has changed on the backend in that code for awhile. We just take whatever is after the `+` verbatim, so I'm pretty sure it's always been that if you upload an SGF with W+R or whatever, it would have always shown "R" as the result. Regardless, this is a good fix / workaround I think, so thanks!

PR #41, opened 2021-10-08
Add a small deadzone in between intersections to prevent misclicks

wmbr:
I noticed that It is easy to misclick if the cursor is at the very edge of the "ghost" stone.
Apparently this also happens to others, e.g. see https://youtu.be/nNxD47Bfhhw?t=1303
This pull request adds a small "deadzone" between intersections, in which clicks have no effect.


anoek:
Thanks, I hope that helps :)

PR #43, opened 2021-11-26
AIReview updates

benjaminpjones:
- Add ability to use score as the metric for "Key Moves"
    - I did it with an additional parameter to allow for score instead of win rate.  Perhaps that puts too much onus on the API clients?  Let me know if I should move more logic into this function.
- Change type definition for SetMarks (I don't think it's actually meant to handle a MarkInterface proper)


anoek:
Awesome, thanks!

Yeah that `setMarks` system doesn't make sense to me as I read it now. I think the mark system needs a bit of love, the labels score stuff has been bolted on in kind of an awkward fashion, and you're right the MarkInterface typing of that only hurt not helped in understanding what it was trying to do much less actually do.

PR #44, opened 2021-11-27
add getWorstMoves

benjaminpjones:
This function is a little smarter than compute worst moves as it determines whether to use score on its own, and also filters out non-blunders.

### DIFF
--- a/src/AIReview.ts
+++ b/src/AIReview.ts
@@ -36,7 +36,7 @@ export interface AIReviewWorstMoveEntry {
  *     determining the worst moves, rather than win rate. This is useful for
  *     handicap games where the winrate is 100% for a large portion of the game.
  */
-export function computeWorstMoves(starting_move:MoveTree,
+function computeWorstMoves(starting_move:MoveTree,
                                   ai_review:JGOFAIReview,
                                   use_score = false):Array<AIReviewWorstMoveEntry> {
     let ret:Array<AIReviewWorstMoveEntry> = [];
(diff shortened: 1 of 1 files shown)


src/AIReview.ts
```diff
@@ -76,3 +76,26 @@ export function computeWorstMoves(starting_move:MoveTree,
...
+        threshhold = -0.2;
+    }
+
+    const filtered_worst_moves = worst_moves.filter(de => de.delta <= threshhold);
+
+    if (filtered_worst_moves.length >= 3) {
+        return filtered_worst_moves;
+    }
+
+    return worst_moves.slice(0, 3);
```
anoek, line 100: Since we're returning a potentially long array `filtered_worst_moves`, seems like we shouldn't be slicing `worst_moves`?

anoek (commented):
Lets remove the `export` from `computeWorstMoves`, it seems like `getWorstMoves` is going to be the function anyone wants to call. Alternatively, rename both to keep the interface consistent, although `getWorstMoves` is probably a better name anyways.

src/AIReview.ts
benjaminpjones: I think this is by design?  So we only return that slice in the event that the filter was too strong (say only 2pt losses the entire game)

benjaminpjones:
Went ahead and removed the export from `computeWorstMoves`.  Left the name as-is as I think `getWorstMoves` is different enough in what it returns.

anoek:
Thanks for the improvement!

PR #45, opened 2021-11-27
Some more typings

### LINKED ISSUE #30: Change GoEnginePlayerEntry.pro to boolean
Request: Change [GoEnginePlayerEntry.pro](https://github.com/online-go/goban/blob/be225ec15a99f65fd5794d52ac75c0f68aa39b74/src/GoEngine.ts#L84) to boolean.

Would this cause issues on the server side of OGS?  Currently, the front end treats it like a bool anyway:

        let pro = ev.target.value.indexOf(".1") > 0;    // result of comparison is boolean
        console.log(rank, pro);
        this.props.config.players.black.rank = rank;
        this.props.config.players.black.pro = pro;
        this.props.black.rank = rank;
        this.props.black.pro = pro;

Many similar examples in [GameInfoModal.tsx](https://github.com/online-go/online-go.com/blob/d0e87b841fef869428ff5b75e77a5f0442251f06/src/views/Game/GameInfoModal.tsx).  I ran into this when trying to add type information, in order to make [#1362](https://github.com/online-go/online-go.com/issues/1362) easier to solve.

benjaminpjones:
Fixes #30 

This fixes some type issues that I found when cleaning up GameInfoModal (PR incoming)

- GoEnginePlayerEntry.pro changed to boolean
- GoEngineConfig.winner added

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -81,7 +81,7 @@ export interface GoEnginePlayerEntry {
 
     /** XXX: The server is using these, the client may or may not be, we need to normalize this */
     name?: string;
-    pro?: 0 | 1;
+    pro?: boolean;
 }
 
 export interface GoEngineConfig {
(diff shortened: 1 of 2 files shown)


anoek:
Thanks!

PR #46, opened 2021-12-17
Add rengo fields to the engine (plus lint fix)

GreenAsJade:


anoek:
Awesome, I think this is fine to merge before rengo is live so i'm gonna go ahead and  do it. It's not clear to me if we need to do something more with decoding that new `played_by` field or not, I reckon it'll sort itself out in beta testing though.

PR #48, opened 2022-01-01
Rengo support - store per-move player_updates in Move Tree and emit for changes

GreenAsJade:


src/GoMath.ts
```diff
@@ -380,6 +385,24 @@ export class GoMath {
         return ret;
     }
 
+    public static stripModeratorOnlyExtraInformation(move: AdHocPackedMove): AdHocPackedMove {
+        const moderator_only_extra_info = ['blur', 'sgf_downloaded_by'];
+
+        if (move.length === 5 && move[4]) { // the packed move has a defined `extra` field that we have to filter
+            let filtered_extra = move[4];
+            for (const field of moderator_only_extra_info) {
+                let {[field]:hide_me, ...more_filtered} = filtered_extra;
```
anoek, line 394: Fancy :)

GreenAsJade: In the end probably silly really - it would be much simpler to just use `delete` inside the loop

I ended up with this because I _really_ wanted to be able to say  

`{...[moderator_only_extras], ...filtered_extra} = move[4]`

which would not need the loop.    I was twisting and turning to try to find a way of saying that, but as you see I never got there, I ended up with this instead :D

anoek: Yeah I guess we're duplicating the object `n` times this way, still cool `[field]:hide_me` syntax though

GreenAsJade: 😁   yep :)      I realise the answer to my question (when would you use it) might be if it were `[field]:use_me_locally`

PR #50, opened 2022-01-16
Add prettier

benjaminpjones:
## Proposed Changes
- Add Prettier
- Add GitHub workflow check
- Run Prettier (twice)
   - Prettier had trouble over a bit of code with an inline comment and needed two passes.  Issue appears to be transient and harmless.

Note: I didn't bother removing formatting rules from TSLint since there wasn't a CI check and lint already fails.

### DIFF
--- a/src/GobanCanvas.ts
+++ b/src/GobanCanvas.ts
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
-
-import { JGOF, JGOFNumericPlayerColor } from './JGOF';
-import { AdHocFormat } from './AdHocFormat';
+import { JGOF, JGOFNumericPlayerColor } from "./JGOF";
+import { AdHocFormat } from "./AdHocFormat";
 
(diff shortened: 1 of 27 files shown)


src/GobanCanvas.ts
```diff
@@ -756,33 +865,45 @@ export class GobanCanvas extends GobanCore  {
...
-                                           }, this.puzzle_autoplace_delay);
-                                       }
+                                    let next =
+                                        this.engine.cur_move.branches[
+                                            getRandomInt(0, this.engine.cur_move.branches.length)
+                                        ];
+
+                                    if (
+                                        calls === 1 &&
+                                        /* only move if it's the "ai" turn.. if we undo we can get into states where we
```
benjaminpjones: Here's the hunk that required two runs.

anoek:
Awesome, thanks @benjaminpjones !

PR #51, opened 2022-01-19
Changed == to === (and removed semicolons)

GreenAsJade:
... in its own commit, in case it breaks something.


anoek:
Thanks for the separation, I *think* that should be a-ok

PR #52, opened 2022-01-19
Take out TSLint rules that look like they do formatting.

benjaminpjones:
I don't know if this is all of them, but this is a quick fix for some prettier conflicts @GreenAsJade bumped into.

Long-term plan is to nix TSLint in favor of ESLint, hence why not a more thorough investigation of prettier compatibility.


anoek:
Awesome sounds good

PR #53, opened 2022-01-19
Eslint

benjaminpjones:
Luckily the tslint files for the React repo and this one were virtually identical (https://www.diffchecker.com/jWct3tgN).  Therefore, I figured I could just use the same .eslintrc file, and just suppress any rules that don't run (yet).

## Proposed changes

- Add ESLint and a bunch of plugins (the same ones in the React repo)
- Copy over the config
- lint --fix pass (mostly `let` -> `const`)
- suppress broken rules
- fix one floating promise error and turn that rule back on

## Not done in this PR

- add lint to CI
- Fix `eqeqeq` errors (#51 addresses those)
- remove tslint config


anoek:
Thanks @benjaminpjones !

PR #54, opened 2022-01-19
Add ESLint to the CI, remove TSLint config

benjaminpjones:
Follow up to #53 

## Proposed changes

- Add lint to the GitHub workflow
- Add a script to run all CI checks (can be used to manually check the lint/formaters before pushing)
- Remove the TSLint config as it is no longer being used


package.json
```diff
@@ -26,7 +26,8 @@
         "lint:fix": "eslint --fix src/ --ext=.ts,.tsx",
         "typedoc": "typedoc --mode file --out doc/ --theme minimal --tsconfig tsconfig.json --excludeExternals --externalPattern '**/node_modules/**' --includeDeclarations src/",
         "prettier": "prettier --write \"src/**/*.{ts,tsx}\"",
-        "prettier:check": "prettier --check \"src/**/*.{ts,tsx}\""
+        "prettier:check": "prettier --check \"src/**/*.{ts,tsx}\"",
+        "checks": "npm run lint && npm run prettier:check"
```
benjaminpjones, line 30: @dsaxton This PR is for Goban, but something like this what you were talking about doing for the React repo?

dsaxton: Indeed, potentially adding TypeScript compilation too if that's seen as useful enough

anoek:
Thanks @benjaminpjones !

package.json
benjaminpjones: Yeah I think that would be good.  I can do the same for the other repo, and add that TS check (I would find that useful!)

PR #56, opened 2022-01-21
make handicap adjustment apply under Japanese rules with the (basic) score estimator, since it uses area counting

pdg137:
This should fix the issue with wrong Japanese scoring:

https://github.com/online-go/online-go.com/issues/1610

We should test a variety of combinations of rules and handicaps before releasing this change!


anoek:
Closing in favor of #57

PR #59, opened 2022-02-12
Add engine.isParticipant()

GreenAsJade:
... so that we have a single way of asking if someone is a participant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "goban",
-    "version": "0.5.75",
+    "version": "0.5.76",
     "description": "",
     "main": "lib/goban.js",
     "types": "lib/goban.d.ts",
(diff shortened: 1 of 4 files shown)


benjaminpjones:
lol look what i found.. 😅 

https://github.com/online-go/goban/blob/c6c9f42d566497a4a6b0247384f759b6b44c0f7c/src/GobanCore.ts#L3485-L3490

GreenAsJade:
🤦‍♀️    Darn!   I searched for isPlayer ...

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "goban",
-    "version": "0.5.75",
+    "version": "0.5.76",
```
anoek, line 3: FTR you don't need to bump this, it happens automatically on release

anoek:
Thanks for the improvement :)

PR #63, opened 2022-03-09
Drawing improvements branch

upsided:
I refactored the GobanCanvas code to allow for faster redraw & prettier graphics capability. Gridlines are moved to a separate layer and drawSquare() draws partial surrounding stones so that stones & shadows can overlap. I also reduced redraw() counts and removed drawingHash, and now the sample count of drawSquare() is below network access in my Performance testing. There's still stuff to do, though...

This was all in preparation for JSONTheme, which was my main goal. It is a class that uses simple JSON to style the goban. This gives OGS Sabaki-like quality to the goban, and allows users/artists to design & share their own themes using basic image URLs & metrics. This could be standard or used as a premium bonus of some kind.

Anyway, using JSON lowers the barrier for artists to contribute to OGS's improvement, since it is data-based, and similar to css. This was my goal.

I have a fork of OGS server that implements a basic UI for experimenting with JSONThemes in the GobanThemePicker via alt-click. I call it "community themes" and will PR that today 
(EDIT: ATM the OGS fork was broken with [660e8c5](https://github.com/online-go/goban/pull/63/commits/660e8c5b76e3089ed982698be22466a2de837148) )


anoek:
That's quite the update :) This will take a good bit of review, but it sounds promising. I did hate the hash system, so I look forward to see what you did to keep performance up while obviating that. It'll probably be a few days before I can do a thorough review and test it in all the places it's used.

In the mean time, there seems to be a graphical glitch on the new lines layer, namely the lines are blury and don't intersect the star points in the center:

image

This is observed on the test board at http://0.0.0.0:9000/ using both firefox and chrome. On mobile, the bluryness isn't there, however the intersection is off.


Also, to fix the lint errors, you can run `npm run prettier`, and it looks like you've obviated some variable names too so those can be removed most likely to fix those errors.

upsided:
Thanks, I fixed the rendering issue and aligned the drawing to tween-pixel boundaries to get crisp lines and hoshi.

src/themes/JSONTheme.ts
```diff
@@ -0,0 +1,1003 @@
...
+Eventually (one hopes) users can supply their own JSON themes & share them
+
+*/
+
+/*  A NOTE
+    So, it's complicated.
+    1. A GoTheme is created as one thing (via "class"), but
+    2. GoThemes are then separated by board & stone rendering, and
+    3. GoThemes are never treated like objects, but passed around as classes that components can instantiate
+    4. Every GoTheme (class) must be hardcoded into GoThemes.ts, and is immutable throughout the duration of app run.
```
anoek, line 38: This may be how it's implemented because I didn't have a use case for anything fancier, but if you wanted to dynamically set those that should be just be fine.

src/themes/JSONTheme.ts
```diff
@@ -0,0 +1,1003 @@
...
+                }
+            }
+        }
+    }
+
+    public rebuildMatrices(seed: number) {
+        this.matrices = JSONTheme.buildMatrices(seed);
+    }
+    protected static buildMatrices(seed: number): Array<MatrixStore> {
+        // have to use static method because theme changes arent honored due to pre-rendered stone caching
```
anoek, line 229: Sounds like the cache needs to be purged?

src/themes/JSONTheme.ts
```diff
@@ -0,0 +1,1003 @@
...
+            }
+            ctx.restore();
+            ctx.setTransform(t);
+        }
+    }
+
+    /* Should return true if you would like the shadow layer to be present. False
+     * speeds up rendering typically */
+    public stoneCastsShadow(radius: number): boolean {
+        return true; // themes will always be given a shadow layer
```
anoek, line 705: Is this a functional necessity? This exists because when we render a lot of mini boards on lower power devices, it made a notable difference.

src/themes/JSONTheme.ts
```diff
@@ -0,0 +1,1003 @@
...
+
+        return json;
+    }
+}
+
+export default function (GoThemes: GoThemesInterface) {
+    JSONTheme.setJSON(JSONTheme.getDefaultJSON());
+    GoThemes["black"]["JSONTheme"] = JSONTheme;
+    GoThemes["white"]["JSONTheme"] = JSONTheme;
+    GoThemes["board"]["JSONTheme"] = JSONTheme;
```
anoek, line 991: Instead of using a static theme system, what about doing something like using a generator to create your instances of JSONTheme classes specific for whatever settings the user has provided? Something like

```
function JSONThemeGenerator(json: string): typeof JSONTheme {
    return class Theme extends JSONTheme {
        public json:any = JSON.parse(json);
    };
}
```

then

```
GoThemes["board"]["hikaru"] = JSONThemeGenerator(hikaru_json);
```
where `hikaru_json` would be that string blob from up above? You could then generate those dynamically and load them localStorage or whatever you wanted and we could avoid the headache of the static settings and whatnot?

src/GobanCanvas.ts
```diff
@@ -1140,54 +1196,61 @@ export class GobanCanvas extends GobanCore {
...
     }
-    public drawSquare(i: number, j: number): void {
-        if (i < 0 || j < 0) {
-            return;
-        }
-        if (this.__draw_state[j][i] !== this.drawingHash(i, j)) {
-            this.__drawSquare(i, j);
-        }
-    }
-    private __drawSquare(i: number, j: number): void {
```
anoek, line 1151: I don't think we can get rid of the drawing hash system, doing some testing on OGS it results in a very notable increase in display latency when you're paging through moves, specifically advancing 10 moves at a time the latency on my machine went from 100ms to about 400. There are likely other use cases where there's degragation too.

Was there a particular challenge you were running into with the hash system? I don't particularly like it, but the benefits of it are pretty apparent so unless we can replace it with something better, we should keep it.

src/themes/JSONTheme.ts
upsided: No, not really if the json theme doesn't have shadows. I think I assumed it was only asked once so I set it true to cover all bases.

Fixed with [4d92fb2](https://github.com/online-go/goban/pull/63/commits/4d92fb2cc3c8f28eae21e5a6dc8eca03cc2cbea2)

src/themes/JSONTheme.ts
upsided: Yeah I think that comment was because I was trying to buildMatrices() in "prerenderStone" which wasn't working. Now I build them in the constructor, so the static method isn't necessary.

Fixed with [b362b53](https://github.com/online-go/goban/pull/63/commits/b362b53bcc7bddcb5ec28fde0c4d0fcb9857b1e8)

src/GobanCanvas.ts
upsided: To get an extended draw area for the square, I draw the surrounding squares to mix them without glitches. This allows nice shadows and effects like jostled stones, etc. and puts Goban on par with Sabaki.

DrawingHash() would skip these needed draws, so it was a challenge.

I do "kind of" hash inside redraw(), using json.stringify(). I compare the board state to its previous state and only draw a square if different. Full redraw()s were the bottleneck so I moved the "hash" idea there. I also figured stringify() would be faster because it's native.

Anyway, the current lag is my fault; I was planning to gather all needed stone draws and calculate a clip at once for "big redraws", instead of drawing squares over and over, but didn't get to it.

Now I have, in [6b60db3](https://github.com/online-go/goban/pull/63/commits/6b60db36fcb2410907ff8159ec0c002e00f9b5d9). It's a pretty big speedup for me, try it out. It's pretty close to ideal (only draws stones that need it), so if it's still slow I'll have to whip out some other strategy like limiting the extended clip box to only shadows or something.

anoek: Awesome, that brought it down to be about as fast as the hash system, it's still maybe 25% slower but it's not that noticeable, and you're doing more so I'm fine with that. Though there's some quirks to work out:

image

image

You can test this if you have your online-go.com repo pointed to the beta (the default) and go to

http://localhost:8080/game/9594

and going to move 36

upsided: labels weren't getting redrawn because I only drew when a stone was there. Now it draws when label or stone exists.
Fixed with [4b2a81c](https://github.com/online-go/goban/pull/63/commits/4b2a81c4ddab87afde9e909874b896943e8bd5df)
Also, faded lines are drawn in the right order and look less intrusive

upsided: I improved it a little by paying attention to theme's bounding box. For simple themes (plain, rendered) the draws are now marginally faster because they don't overlap like a JSON theme.
I also put the last move marker in its own little element so as not to keep drawing stones beneath it. 
Finally, JSON themes won't draw shadows when radius < 10. Took me a day + some tea to realize what you meant, sry.
This is in [61b07ca](https://github.com/online-go/goban/pull/63/commits/61b07ca8d207c42da7560e9a08d1554be6985eb6)

src/themes/JSONTheme.ts
upsided: Yeah, I like this. I was afraid of integration with OGS but I'm pumped up enough to send a legit PR that implements it. If JSONTheme is instance-based, then the user can have all kinds of themes at their beck & call. Was thinking a "horizontal scroll" theme picker with a "+" on the end to add another...

anoek:
Awesome, I'm not seeing those glitches anymore. It slowed down a little, we're now at about half as fast as we were before. I think though we might optimize things on the ogs side a little too, I think the nav +- 10 code is forcing a redraw each iteration which is somewhat silly, so if we defer that and just render it once it'd probably make things feel snappy again.

Are you wanting to do the JSON class instancing stuff in this PR or a future PR?

upsided:
I did the instancing. Now it's possible to pass json themes to construct when asking for a GoTheme list.

ATM, I've put in some stock themes (hikaru, etc) that are included when no json is passed in. This is nice for testing but it does reference external websites. Maybe they should be removed or put in assets if they're nice enough to keep

upsided:
I had some pretty terrible math that filtered the stones to draw; I must have been seeing ghosts when I tested. Now it's proper and should be faster on built-in themes.

If you're willing to merge I should mention is that it adds about 6 themes to the stock themes and it doesn't fully work with my "community theme" OGS fork (no big loss, tbh). Also, URL images do not load on the preview in OGS due to network latency; Putting onload()s for the images was something i wanted to get to eventually.

upsided:
Created a cached/prescaled image class for JSONThemes; now images draw at destination size via ImageBitmap. I didn't notice much of a speed diff on my comp, but the advantage is that the class that pre-scales images also attempts lazy loading, so that OGS (as it's now written) will be more likely to see the images in the picker.

Since they're destination size now, it might be faster to set the stone canvas image-rendering to "pixelated", but I dunno

anoek:
Hm, yeah it seemed to hurt if anything. I have an injury that makes it a bit hard to do real work at the moment, I'm hoping I'll be back in the swing of things mostly next week but we'll see. When it does I can delve in and try and hunt down the performance issue, I don't see any reason offhand why this should be slower than the pre-rendered stones but it still is.

The other thing we still need to fix before we merge is the deferred rendering stuff so we don't get blank boards and missing stones in the theme preview, which we're still getting. For that I think we basically just need to call `place(Black|White)Stone` again when all of the images have finished loading, but how to structure that is an open question.. might make `GoTheme` an event emitter we subscribe to the `load` event or something, or just stuff a callback somewhere, not sure the best way off the top of my head. Whatever we pick should work with the main Goban too so we can trigger a re-render when everything is loaded I reckon.

anoek:
As I was working through some other stuff I noticed there's a shadow popping thing going on sometimes, I took a pair of screenshots.. they look really similar at a glance but you'll notice in the second the shadow is a little darker / bigger, this happens as I was moving my cursor stone around the board, all the shadows in surrounding areas were popping a bit

image

image

anoek:
Also these green squares aren't getting cleared correctly FYI, if we mouse over them they clear correctly, just not as we skip around in the game.

image

upsided:
Sorry to hear about your injury, hope you get well soon!

I fixed the shadow issue by abandoning my brilliant (?) ImageBitmap idea. I'll account for the heatmap when checking if a square changed. 

I still need to play around with the deferred stuff; it's no trouble to redraw the board itself (calback, event, whatever), but the way the OGS picker calls PlaceXStone() directly led me to trying a timeout scheme that works for me now but is still *suspicious*.

Also in the works is fixing the ugly in the themes themselves and maybe convering them to actual .json files instead of the hardcode going on.

I won't be updating till next week anyway. So please chill & get better!

upsided:
The theme picker wasn't drawing correctly with deferred draw for a couple reasons. One was that I didn't keep track of render context state, which would change between timeouts(), another was that OGS ThemePicker called placeXStone() using the same render ctx for both shadow & stone. This would cause the shadow to draw "over" the stone if it loaded later and thus erase it.  I fixed both issues along with the heatmap lingering bug

Finally, I added images to the assets folder and tweaked the themes so they look nicer.

anoek:
Awesome, those previews are showing up, thanks for your continued work :)

I noticed while clicking around in a review a partial blue circle showing up in the upper left sometimes:

image

This doesn't happen on `main`

Another thing I noticed is one time when I refreshed the board redrew several times with increasingly dark shadows, like it was redrawing every load without clearing or something, and then redrew everything again correctly.. that's not that big of a deal because it corrected itself, but thought I'd note it.

anoek:
Not sure how to reproduce this but FYI I hit this this morning:

image

anoek:
I'm wondering if maybe the thing to do would be to split out this PR a bit? Normally something like this would be split into patches that add the new core functionality required (adding a new lines layer for instance), then adding the new desired functionality (the theme code), then the rendering modifications that were made in the name of improving the rendering system. Keeping things a bit more separate helps isolate where problems like the above might be coming from, and helps keep the changes a bit easier to review in general.

anoek:
Feel free to re-open this if you want to pick it back up. Smaller PR's might be easier to get merged in a timely fashion though.

PR #65, opened 2022-03-31
Tell clients who played a move that was clicked on

GreenAsJade:
The server sends played_by for rengo moves.

This change takes advantage of that to emit a played-by-clicked event when a person clicks on a move with this information in the move tree.

Which in turn lets clients tell the user who played that move.

### DIFF
--- a/src/GobanCanvas.ts
+++ b/src/GobanCanvas.ts
@@ -3132,6 +3132,18 @@ export class GobanCanvas extends GobanCore {
                         this.syncReviewMove();
                         this.redraw();
                     }
+                    if (this.engine.cur_move.played_by) {
+                        // note that getRelativeEventPosition handles various
+                        // nasty looking things to do with Touch etc, so using it here
+                        // gets around that kind of thing, even though in theory it
+                        // might be nicer to sent the client absolute coods, maybe.
+                        const rpos = getRelativeEventPosition(event);
+                        this.emit("played-by-click", {
+                            player_id: this.engine.cur_move.played_by,
+                            x: rpos.x,
+                            y: rpos.y,
+                        });
+                    }
                 }
             } catch (e) {
                 console.error(e);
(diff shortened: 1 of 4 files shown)


src/GobanCanvas.ts
```diff
@@ -3132,6 +3132,18 @@ export class GobanCanvas extends GobanCore {
...
+                    if (this.engine.cur_move.played_by) {
+                        // note that getRelativeEventPosition handles various
+                        // nasty looking things to do with Touch etc, so using it here
+                        // gets around that kind of thing, even though in theory it
+                        // might be nicer to sent the client absolute coods, maybe.
+                        const rpos = getRelativeEventPosition(event);
+                        this.emit("played-by-click", {
+                            player_id: this.engine.cur_move.played_by,
+                            x: rpos.x,
+                            y: rpos.y,
```
anoek, line 3144: `rpos` should be `pos` shouldn't it?

GreenAsJade: I wished it was, so the second  `getRelativeEventPosition` wouldn't be needed.

I grappled with that and found my grasp of the zillions of coordinate reference points lacking :D

I _think_ that we want to tell the client precisely  the result of `getRelativeEventPosition()`.

I'm not completely clear what is happening to pos with  `pos.x += ox;` , but it seemed to be allowing for scrolling to calculate a node-at-location, rather than a physical offset needed to apply to get to the right place on the screen.

Some evidence for this is the card pops up in the right place, though obviously this is only a couple of data points.

Full disclosure: I hadn't actually tested a scrolled pane, but I have now, and it seems to work properly.

anoek: Yeah I think I probably should have finished my tea, lgtm :)

GreenAsJade: :D  Any person who can read this DOM coordinate kind of stuff and know what's right or wrong without trial and error is in the League of the Gods, tea or no tea!

PR #73, opened 2022-05-24
Fix small boards issue

benjaminpjones:
Bug: https://forums.online-go.com/t/new-bug-for-smaller-boards/43299/2

Error arose in https://github.com/online-go/online-go.com/pull/1873

I think there were ~two~ issues about the "load" function:

- square size was not changed if config did not specify a display size.
   - Naturally socket does not send a display size with the "game_data" event
   - this.display_size may be set prior to the load event, in which case we will still need to recalculate the square size
- ~redraw not triggered if engine was previously undefined.~ I read the code wrong, redraw logic was already correct.

### DIFF
--- a/src/GobanCore.ts
+++ b/src/GobanCore.ts
@@ -2044,11 +2044,11 @@ export abstract class GobanCore extends TypedEventEmitter<Events> {
             this.bounded_height = this.height;
         }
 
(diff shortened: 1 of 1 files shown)


anoek:
Thanks for the fix!

PR #74, opened 2022-05-30
Use MA instead of XX for SGF exports.

benjaminpjones:
MA is the officially supported property for marking points with an X: https://www.red-bean.com/sgf/properties.html#MA

This pull request aims to maintain backwards compatibility.  Newly exported SGFs will use the MA field, but the XX field will still be recognized when reading a "legacy" SGF in.

## Additional Context

https://forums.online-go.com/t/ogs-seems-to-not-respect-the-sgf-standard-with-the-x-marking/43361/2


anoek:
Awesome thanks :)

PR #77, opened 2022-08-19
[cleanup] Remove computeProbabl* functions.

benjaminpjones:
I ctrl-F'ed the three repos (react, goban, node) for `computeProb` and couldn't find anything except the definition of these methods.


anoek:
Ah indeed, I think this was part of the auto-scoring stuff i had before we switched to the probabilistic score estimator system, so very ancient by this point.

Thanks for the cleanup!

PR #79, opened 2022-08-22
TestGoban

benjaminpjones:
I tried `new GobanCore` but it expects a bunch of functions to be implemented.  So I made a `TestGoban` instead that I can `new` in tests when I don't need all of `GobanCanvas`.  In the future, utilities might be added to help inspect the internals of GobanCore (e.g. making protected fields public, adding functions like `current_message`) when there isn't a view to hold this state.


anoek:
Awesome, yeah that seems like the best approach, thanks!

PR #80, opened 2022-08-23
goban.pass() to play a conditional move in "conditional" mode

benjaminpjones:
Goban isn't currently used this way, but consider the following code:

```
goban.setMode("conditional");
goban.pass();
```

I believe this should logically result in a pass (`".."`) being appended to the conditional move tree.  Current behavior is that the players get swapped, but no move is added to the conditional tree.  This leads to the possibility that the move tree could be appending black, but the board would show white and vice versa.

<details> <summary> Screen recording: Current </summary>
</img></details>

<details> <summary> Screen recording: Proposed </summary>
</img>
</details>

Also in this PR: test coverage for conditional moves

### DIFF
--- a/src/GobanCore.ts
+++ b/src/GobanCore.ts
@@ -2710,7 +2710,6 @@ export abstract class GobanCore extends TypedEventEmitter<Events> {
     public pass(): void {
         if (this.mode === "conditional") {
             this.followConditionalSegment(-1, -1);
-            return;
         }
 
         this.engine.place(-1, -1);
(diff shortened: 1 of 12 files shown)


src/GobanCore.ts
```diff
@@ -2708,6 +2708,11 @@ export abstract class GobanCore extends TypedEventEmitter<Events> {
         });
     }
     public pass(): void {
+        if (this.mode === "conditional") {
```
benjaminpjones, line 2711: This is the extent of the functional changes.  The rest was just tests I wrote to help me understand the code.

anoek:
Good job tracing that spaghetti mess, you'll make some folks pretty happy by adding support for conditional passes :)

PR #83, opened 2022-09-10
Fix audio-clock color

turtoise0:
Changes:

- Set the audio clock color to the "opposite" color of the last official move instead of `engine.colorToMove()`.

Fixes https://github.com/online-go/online-go.com/issues/1231 and https://github.com/online-go/online-go.com/issues/1723.


anoek:
Thanks for digging into that and getting a fix made!

PR #84, opened 2022-10-01
Fix disable_analysis setting in rengo games

turtoise0:
Fixes https://github.com/online-go/online-go.com/issues/2016

Changes: don't enable analysis if the player is a rengo participant


anoek:
Thanks for the fix!

PR #86, opened 2022-10-05
Fix sgf escaping & minor refactor

jmoudrik:
Core of the PR is fixing SGF escaping issue:

https://github.com/online-go/online-go.com/issues/1547

The SGF changes were not tested, as the SGF is generated server side, but I think it is pretty safe, lints, etc were fine & I triple checked. 

Apart from that, various minor refactors around it were made:
- added tests for SGF escaping
- sgf now uses linear `string` concatenation (via `array.join`) instead of quadratic `txt += "...."`
- refactored coordinate string, to remove copy-paste & improve safety & consistency (moved to `GoMath`)
  - the code was well tested, so pretty safe I hope
- README now has info about tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,3 +51,23 @@ lrwxrwxrwx 1 root root 11 Oct  4 13:10 node_modules/goban -> ../../goban
 $ make
 ```
 
+
+# Running & Writing tests
+Is easy.
+
+## Deps
+Just install `jest` package (might need `sudo`):
+
+```
+$ npm install --save-dev jest
+```
+
+## Running and writing tests
+Tests live in `src/__tests__` directory, check it out & add Your own!
+To run tests:
+```
+# from root directory
+$ cd goban 
+$ npm test
+```
+
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
+
 # Current state
 
 *Warning* This code base has been extracted from the online-go.com source code
@@ -8,6 +9,14 @@ some moderinization of the rest of the site. Until v 1.0 is released of this
 library, it is not recommended to make use of it as much overhaul is happening
 to improve usability on sites other than online-go.com.
 
+
+# Before PR
+
+- be sure you are able to replicate / test what you are fixing :-)
+- write tests for your code if possible, see the test section below
+- be ready for CI check on PR:
+	- run tests `npm test`
+	- run prettify `npm prettier`
 
 # Dev setup
 
@@ -70,4 +79,5 @@ To run tests:
 $ cd goban 
 $ npm test
 ```
+
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ to improve usability on sites other than online-go.com.
 - write tests for your code if possible, see the test section below
 - be ready for CI check on PR:
 	- run tests `npm test`
-	- run prettify `npm prettier`
+	- run prettify `npm run prettier`
 
 # Dev setup
 
@@ -40,9 +40,9 @@ You need to compile goban:
 ```
 $ cd goban
 
-# dependencies (your mileage may vary based on what you have, or have not installed already)
+# dependencies
 $ sudo npm install --global yarn
-$ yarn add webpack-cli --dev
+$ yarn install
 
 # build the goban module
 $ make
--- a/src/GoMath.ts
+++ b/src/GoMath.ts
@@ -283,7 +283,7 @@ export class GoMath {
                     if (mv instanceof Array && typeof mv[0] === "number") {
                         ret.push(decodeSingleMoveArray(mv as [number, number, number, number]));
                     } else {
-                        console.error("mv: ", mv);
+                        //console.error("mv: ", mv);
                         throw new Error(`Unrecognized move format: ${mv}`);
                     }
                 }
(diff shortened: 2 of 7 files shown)


src/GoMath.ts
```diff
@@ -211,9 +211,29 @@ export class GoMath {
...
+
+    public static coor_ch2num(ch: string): number {
+        return GoMath.COOR_SEQ.indexOf(ch?.toLowerCase());
+    }
+
+    public static coor_num2ch(coor: number): string {
+        return GoMath.COOR_SEQ[coor];
+    }
+
+    public static readonly PRETTY_COOR_SEQ = "abcdefghjklmnopqrstuvwxyz";
```
benjaminpjones, line 224: I wonder if this PRETTY_COOR_SEQ should just be uppercase.  Both instances of `pretty_coor_num2ch` is followed by a .toUpperCase(), so I think the assumption is that pretty == uppercase (and no `I`).

jmoudrik:
@benjaminpjones  - thx, super interesting, times they are changin' !
I am tempted to leave it like this anyway, there is some benefit in the `const` probably. Wdyt?

benjaminpjones (commented):
Really nice stuff!  I'm trying to wrap my head around the problem - so I gather the before and after is something like this?

- before: `[[]` --adding escapes-> `\[`
- after: `[[]` --adding escapes-> `[[\]`

Edit: oh I see the GH issue, thanks for the in depth report.

src/__tests__/Misc.test.ts
```diff
@@ -0,0 +1,18 @@
+import { escapeSGFText } from "../Misc";
+
+test("escapeSGFText duplicates slashes", () => {
+    expect(escapeSGFText("f\\*\\a")).toBe("f\\\\*\\\\a");
```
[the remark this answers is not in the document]
jmoudrik: I am at least gonna add a note that half of the slashes are for js input...

README.md
```diff
@@ -51,3 +60,24 @@ lrwxrwxrwx 1 root root 11 Oct  4 13:10 node_modules/goban -> ../../goban
 $ make
 ```
 
+
+# Running & Writing tests
+Is easy.
+
+## Deps
+Just install `jest` package (might need `sudo`):
```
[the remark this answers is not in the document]
jmoudrik: huh, I most definitely did not run `yarn install` myself, nor is it a step in makefile; so I am gonna add the `yarn install` step to README to the setup part, this will remove need for installing the `webpack` dep as well

src/GoMath.ts
jmoudrik: the feeling here was for both char seqs to be the same kind of "source of the truth", and upper/lowercase is just output format, but whatever & we will save an uppercase call & simplify, so dunno; no strong opinion

jmoudrik:
the output is the same anyway, but I think adding a few tests makes sense here anyway, there is more sgf issues in the tracker, so I think this can definitely be hardened

jmoudrik:
@benjaminpjones  Regarding the before & after.

Before: 
```
// duplicate FIRST slash (big bug)
txt.replace(/[\\]/, "\\\\")
// escape all closing brackets (correct)
.replace(/\]/g, "\\]")
// escape all opening brackets (not needed, but fine)
.replace(/[[]/g, "\\[")
```

jmoudrik:
I also extended the `escapeSGF` to handle whitespace as per SGF spec (all non `\n\r` whitespace gets converted to simple space) and optional escape of colon (for obscure properties, such as move COOR ranges, etc (unused now, but at least the `escapeSGF` is fully spec-compliant now).

src/GoMath.ts
jmoudrik: fixed to your tip, no need to bikeshed here, it is a detail

src/__tests__/Misc.test.ts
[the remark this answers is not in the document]
jmoudrik: added `String.raw` to tests

src/GoMath.ts
benjaminpjones: Indeed, I definitely should have labeled this a "nit". Anyway thanks for addressing!

README.md
```diff
@@ -48,24 +39,21 @@ From the `online-go.com` directory run
 
 `yarn link goban`
 
+Once done, your online-go.com development environment will use your development
+`goban` code.
 
-# Running & Writing tests
-Is easy.
```
benjaminpjones: 😅 funny deletion

src/__tests__/GoEngine_sgf.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+function rmNewlines(txt: string): string {
+    return txt.replace(/[\n\r]/g, "");
+}
+/*
+ * check that parse -> serialize roundtrip generates the same sgf
+ * (at least for the moves)
+ */
+test("sgf -> parseSGF() -> toSGF() roundtrip (moves only)", () => {
+    Object.entries(SGF_TEMPLATES).forEach(([key, template], index) => {
```
benjaminpjones, line 34: Nit: you may be interested in [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) which is made for this type of test-cases-in-array pattern.  See an example in @GreenAsJade's PR: https://github.com/online-go/goban/blob/8666815d12dbef1d98c2a6dcf0d3d450cce026c0/src/__tests__/GoMath_positionId.test.ts#L145-L147

benjaminpjones (approved):
Seriously nice stuff in here, I appreciate the extra effort!

src/__tests__/GoEngine_sgf.test.ts
jmoudrik: nice! fixed

anoek:
Awesome, thanks for digging in and putting this together, looks great!

PR #88, opened 2022-10-08
Added `currentPositionId()` to `GoEngine`, calculate this in `GoMath`

GreenAsJade:
This adds a function that provides a unique id for each unique board position.

The first use-case for this is "joseki duplicate position detection".

"Game position search" also seems on the cards, except that the ids for every position in game (even every "interesting" game for some interesting subset) might be a lot of (string) data.


GreenAsJade:
This uses the @benjaminpjones  / Mr Alex recommended "simple" id method, rather than the @benjaminpjones `funky baduk id`, because although the latter generates smaller ids, those guys reckon it's not worth it.

BHydden:
Does this recognise mirrors and rotations as like positions? Or does it treat them as unique? 

(For instance, for a board with only one stone on it, are D4 D16 Q4 and Q16 one board position or four?l

GreenAsJade:
It covers 8 rotations & reflections I can think of :)

BHydden:
That's awesome ❤️

anoek:
Very cool! I'd love to hear from @benjaminpjones as he usually has good insight on these things.

I'm not personally too caring about saving a few bytes here or there in general so this might be moot, but I would note that by sorting strictly lexically like we're doing, we're pretty much always going to be picking an orientation in the upper left as opposed to say the bottom left or bottom right, and it'll probably mean that every string is a few bytes longer. I doubt it'll ever make much of a difference, but if we want to be picky there we might sort by string length then lexically?

benjaminpjones:
> I'm not personally too caring about saving a few bytes here or there in general so this might be moot, but I would note that by sorting strictly lexically like we're doing, we're pretty much always going to be picking an orientation in the upper left as opposed to say the bottom left or bottom right, and it'll probably mean that every string is a few bytes longer.

If we switch to SGF coords ("K10" vs "jj") this wouldn't be an issue, because all moves are the same length right?

A couple more thoughts:

- Don't prefix every move with "B" or "W" - it's implied by the position (before the delimiter == white, after the delimiter == black).  This will save about 33% on space.
- Do swapped colors count as a symmetry?

src/GoMath.ts
```diff
@@ -514,4 +514,168 @@ export class GoMath {
...
+        }
+        let this_way = `${white_state}.${black_state}`;
+        let id = this_way;
+        // console.log("calc, id:", this_way, id);
+
+        // Take care of identical positions that are rotated/reflected: calculate
+        // using the same algorithm for each of the 8 transformations, and chose the
+        // alphanumerically shortest version as the actual id...
+
+        black_state = "";
```
benjaminpjones, line 557: I hate to be a DRY zealot, but I think we're talking 150 lines vs 50 lines here.

One approach to reduce duplicate code is to define the coordinate transformations at the outset and loop:

```
tforms = [
    (x, y) => ({x, y}),
    (x, y) => ({x, y: height - y - 1})
        ...
    (x, y) => ({x: y, y: width - x - 1})
    (x, y) => ({x: height - y - 1, y: width - x - 1})
]

ids = []
for (tform of tforms) {
    let black_state = "";
    let white_state = "";
    for (let x = 0; x < width; i++) {
        for (let y = 0; y < height; y++) {
            const c = tform(x, y);
            if (position[row][col] === JGOFNumericPlayerColor.BLACK) {
                black_state += encodeMoves(c.x, c.y)
            }
            if (position[row][col] === JGOFNumericPlayerColor.WHITE) {
                white_state += encodeMoves(c.x, c.y)
            }
        }
    }
    ids.push(`${blck_state}.${white_state}`)
}
// sort and return the ids
```

(one benefit of `encodeMoves` vs `prettyCoords` is it doesn't require a height/width!)

anoek:
> If we switch to SGF coords ("K10" vs "jj") this wouldn't be an issue, because all moves are the same length right?

Yep true
 
> A couple more thoughts:
> 
>     * Don't prefix every move with "B" or "W" - it's implied by the position (before the delimiter == white, after the delimiter == black).  This will save about 33% on space.

Indeed

> 
>     * Do swapped colors count as a symmetry?

It seems like for joseki we'd want that to be true, though it's not immediately clear to me whether that should be done as a 2x query kind of thing, or with the position id.

BHydden:
Is B and W still implied in handicap games with x stones that need recording? Isn't it then not a simple matter of 2n=white 2n+1=black? Or am I missing something?

src/GoMath.ts
```diff
@@ -514,4 +514,168 @@ export class GoMath {
...
+        height: number,
+        width: number,
+    ): string {
+        // The basic algorithm: list where the stones are, in a long string :)
+        let black_state = "";
+        let white_state = "";
+        for (let row = 0; row < height; row++) {
+            for (let col = 0; col < width; col++) {
+                if (position[row][col] === JGOFNumericPlayerColor.BLACK) {
+                    black_state += "B" + this.prettyCoords(row, col, height);
```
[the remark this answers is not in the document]
GreenAsJade: Thanks ... I did benefit from pretty coords during debug, but I think the space saving is worth the extra mental effort in the future, now that it basically (AFAICT) works.

src/GoMath.ts
GreenAsJade: Yeah, once again, thanks ... the version submitted is the "make it really easy to understand" version: I consciously decided to dupe rather than loop for that purpose, but your version is elegant.

GreenAsJade:
As it happens, for joseki, we don't need to worry about B/W symmetry, because joseki editors always play black first.

However, it's definitely a feature for position search.   I wonder whether it's simple a case of swapping the before and after the colour delimiter, and chosing the lowest of those two?    I didn't "think this through", it's just a hunch, but it seems logical?

src/GoMath.ts
GreenAsJade: Is there some reason not to just keep the lowest id, and update as we go?

benjaminpjones: Not really, no.  I think I wrote it that way because I actually thought about writing this with a more functional loop:

```
const ids = tforms.map((tform) => {
    /* calculate id for this transform */
    return id;
}
```

(I prefer `map`, but thought maybe for loops are more familiar to other folks?)

GreenAsJade: Yeah - I went through a phase of using functional loops all over the place, but found that unless you have a "good reason" (such as a set of operations on a stream) they can be harder to debug - sometimes because the reasoning is a little more effort, and sometimes just because sometime inserting debug statements is a hassle.

BHydden:
I like the built in tests 👍

src/GoMath.ts
```diff
@@ -533,149 +539,44 @@ export class GoMath {
...
+                    }
+                    if (position[x][y] === JGOFNumericPlayerColor.WHITE) {
+                        white_state += GoMath.encodeMove(c.x, c.y);
+                    }
                 }
             }
-        }
+            const this_id =
+                black_state < white_state
+                    ? `${black_state}:${white_state}`
```
benjaminpjones, line 574: Note on the colon:

I realize the current use case is not URLs, but `:` isn't as URL-safe as other symbols (`-._~`).  It's not hard to address later (using `replace()` or URL escaping), but I thought it's worth bringing up now in case an unreserved delimiter would be preferable.

src/__tests__/GoMath_positionId.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            [0, 0, 1, 0, 0, 0, 0],
+            [0, 0, 0, 2, 0, 0, 0],
+            [0, 0, 0, 0, 0, 0, 1],
+        ],
+        id: "abdgfd:bccfeega",
+    },
+] as Array<Testcase>;
+
+test("Position IDs", () => {
+    TEST_BOARDS.forEach((testcase: Testcase) => {
```
benjaminpjones, line 176: Nit: You might be interested in [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout)

Basically made for this use case, but might give better error messages on test failure.

benjaminpjones (approved):
Tests look good!

Re: Colors, one thing to think about is that there are Joseki positions that look the same color-flipped, but they differ by who plays next.

e.g.

- [3-4, one space approach](https://online-go.com/joseki/26969)
- [5-4, 3-4 invasion](https://online-go.com/joseki/30755)

"Tenuki" could also cause similar issues (although I didn't find any tenuki variations as examples)

At any rate, I don't see that as being an issue with this PR, just something that isn't apparent we're handling it yet.

GreenAsJade:
> Tests look good!
> 
> Re: Colors, one thing to think about is that there are Joseki positions that look the same color-flipped, but they differ by who plays next.
> 
> e.g.
> 
> * [3-4, one space approach](https://online-go.com/joseki/26969)
> * [5-4, 3-4 invasion](https://online-go.com/joseki/30755)
> 
> "Tenuki" could also cause similar issues (although I didn't find any tenuki variations as examples)
> 
> At any rate, I don't see that as being an issue with this PR, just something that isn't apparent we're handling it yet.

Dang, that does seem like an issue for joseki.   I think you've answered the question "why don't we implement colour symmetry?".   The answer is "because convention is that black plays first".

https://online-go.com/joseki/29105  is a _different_ position to https://online-go.com/joseki/26969 from OJE's point of view.

The nice thing about the current position id format is that it is extremely easy to ask "what about colour symmetry equivalence" given the position ids of two positions.     So we have no reason to embedd that check into the position id itself.

src/__tests__/GoMath_positionId.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            [0, 0, 0, 0, 1, 0, 0],
+            [0, 1, 0, 0, 0, 0, 0],
+            [2, 0, 0, 0, 0, 0, 0],
+            [0, 0, 1, 0, 0, 0, 0],
+            [0, 0, 0, 2, 0, 0, 0],
+            [0, 0, 0, 0, 0, 0, 1],
+        ],
+        id: "abdgfd:bccfeega",
+    },
+] as Array<Testcase>;
```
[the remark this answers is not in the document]
GreenAsJade: Thanks, I do know how that ended up that way, but it's not a good reason :)

src/GoMath.ts
GreenAsJade: I wondered why you used `.`   Good reason!

GreenAsJade: I wish I could say why all of `-.,~` feel inferior as a separator to `:`, but OTOH maybe that's why `:` is the separator in URLs 😄

anoek:
Awesome, thanks for the work! I'm merging, with the understanding that if your needs change while doing the joseki ui side of things we'll just modify this as needed until things are as how you want them, then it'll kinda solidify to be our standard immutable position id.

Issue #92, opened 2022-11-07
Right-clicking should not place a stone.


wmbr:
See https://forums.online-go.com/t/add-option-to-disable-right-click-placement/45422 and https://forums.online-go.com/t/wish-no-right-click-for-moves/38862

PR #93, opened 2022-11-07
Do not place stones on right mouse clicks.

wmbr:
This disables right-click stone placement, fixing https://github.com/online-go/goban/issues/92

### DIFF
--- a/src/GobanCanvas.ts
+++ b/src/GobanCanvas.ts
@@ -697,6 +697,9 @@ export class GobanCanvas extends GobanCore {
         ) {
             return;
         }
+        if (right_click) {
+            return; // we do not have any actions bound to right clicks
+        }
 
         const pos = getRelativeEventPosition(event);
         const xx = pos.x;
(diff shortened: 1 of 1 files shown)


benjaminpjones:
> Can you fix this up so that it only affects games?

> I don't immediately see a way of making the change without either leaving the function even worse than it already is or doing some major refactoring

Wouldn't something like this do it?

```
if (right_click && this.mode === "play") {
    return;
}
```

FWIW, I have no argument with the idea that this method could benefit from test coverage/refactoring, but that is a heftier project, complicated by the fact that we'd need to simulate canvas events.

anoek:
I reckon so, and I'm not necessarily opposed to changing how things work for puzzles, but we'd need to replace that functionality I reckon and probably draw in the community a little.. so something that restricts the change in behavior to just the play phase would be an easier merge for sure.

wmbr:
> Wouldn't something like this do it?
> 
> ```
> if (right_click && this.mode === "play") {
>     return;
> }
> ```

That looks like a practical solution. Alternatively:
```
if (right_click && this.mode !== "puzzle") 
```
(Not sure in which modes "right click acts as left click" is intended behavior).

PR #96, opened 2022-11-19
GobanCanvas tests

benjaminpjones:
Prompted by discussion in #93, this PR has a few basic tests on GobanCanvas that interact with a simulated canvas.

The actual test coverage in this PR is minimal, but hopefully this reduces the "activation energy" required to write unit tests.

- test a stone is placed in play mode
- test a stone is not placed without enabling stone placement
- test no stone is placed if clicked in the "dead zone between intersections


anoek:
Thanks!

PR #97, opened 2022-11-20
Remove pattern search mode

benjaminpjones:
So I was looking into `onTap()`, and I noticed there was this `"pattern search"` mode.  A couple reasons why I think it might be okay to remove it:

- `pattern_search_color` only ever gets set to `0`, so some of this code is flat-out dead
- The other repos don't have any mentions of "pattern search"

I could totally be missing something, but if not, I think this is a nice opportunity to simplify the Goban code!

### DIFF
--- a/src/GobanCore.ts
+++ b/src/GobanCore.ts
@@ -87,8 +87,7 @@ export type GobanModes =
     | "analyze"
     | "conditional"
     | "setup"
-    | "edit"
-    | "pattern search";
+    | "edit";
 
 export type AnalysisTool = "stone" | "draw" | "label";
 export type AnalysisSubTool =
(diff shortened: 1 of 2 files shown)


src/GobanCore.ts
```diff
@@ -87,8 +87,7 @@ export type GobanModes =
     | "analyze"
     | "conditional"
     | "setup"
```
benjaminpjones, line 89: It seems like "setup" and "edit" are similarly unused.  There is no logic attached to "setup", so I think this may have been due to confusion with the puzzle_mode setup. "edit" does have logic, but is unused in the other repos.

anoek:
Yeah pattern search should be safe to remove, we used to support pattern searching games but it was kinda quirky and expensive to support, and rarely used so we removed it. I did a quick search and I think you're probably right about setup/edit too, I though those were used for puzzle stuff but it doesn't look like it, at least not anymore.

PR #98, opened 2022-11-27
Prevent right-click from placing a stone in play-mode

### LINKED ISSUE #92: Right-clicking should not place a stone.

GreenAsJade:
Fixed #92.

This is the change suggested by @benjaminpjones  in #93 - the minimal change to deal with the specific pain point.

It does not attempt to make right-click consistent across modes, it just stops it placing a stone in play mode.

### DIFF
--- a/src/GobanCanvas.ts
+++ b/src/GobanCanvas.ts
@@ -696,6 +696,12 @@ export class GobanCanvas extends GobanCore {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


anoek:
Thanks!

PR #99, opened 2022-11-28
Hodge-podge of small code health improvements

benjaminpjones:
See discussion on #97 for context

- Remove "edit" mode
    - It was unclear what this mode was used for.  It seems similar to placing white or black stones in "analyze" mode, except the "game/move" message is emitted.  At any rate, neither of the other repos use this mode.
- Remove "setup" mode
    - No logic associated with this mode. Likely some confusion with `PuzzlePlacementSetting.mode`.
- Add test for the label subtool (the test places one "triangle" label)
    - Technically, this is not handled in `onTap()`, but we do need to make sure `onTap()` exits without unwanted side-effects
- Add a test for one click submit, mocking the socket
    - I also added a type for `server_socket` in `GobanConfig`.  Does not appear to break the other repos.
- Delete a duplicate test - somehow I ended up with two  "shift clicking in analyze mode jumps to move" test in a previous PR.  I got rid of one. 
- Increase lines covered to 45% (I think just importing `GobanCore`/`GobanCanvas` accounts for most of this)

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -44,7 +44,7 @@ export default {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     "global": {
-      "lines": 4
+      "lines": 45
     }
   },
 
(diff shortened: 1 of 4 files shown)


anoek:
Awesome, thanks for the cleanup!

PR #100, opened 2022-11-29
Remove submitBlinkTimer.

benjaminpjones:
One would expect that this field would be set with the result of a `setTimeout()`, but it never is.  The property is protected,  and as far as I know, `GobanCanvas` is the only relevant class to inherit from `GobanCore`, so if `GobanCanvas` is not setting the timeout either, it is safe to remove.


anoek:
Yep, some old dead code there, thanks for the cleanup!

PR #101, opened 2022-12-04
More onTap() tests

benjaminpjones:
- Calling submit_move() too quickly (<50ms) results in no submission
- Calling submit_move() submits a move
- Right clicking in play mode has no effect (See #98)
- Clicking in empty space during stone removal submits removal message
- Shift-clicking during stone removal toggles single intersection
- Clicking stones during stone removal sends two removal messages
    - This doesn't seem necessary, but I test for it anyway to document current behavior.  I left a note that the test may change if we change the behavior.
- Clicking in scoring mode triggers score estimation handler
     - this technically doesn't touch `onTap()`, but I included it because it illuminates [a dead codepath](https://github.com/online-go/goban/blob/6abfb7277012f4d096baf2ce5cdb1280ee279aa6/src/GobanCanvas.ts#L819-L825) that I'd like to remove. 

Side note: you can find a pretty coverage report in the `coverage/` directory after running tests:

```
yarn test
open coverage/lcov-report/src/GobanCanvas.ts.html
```

### DIFF
--- a/src/__tests__/GobanCanvas.test.ts
+++ b/src/__tests__/GobanCanvas.test.ts
@@ -4,7 +4,7 @@
 
 (global as any).CLIENT = true;
 
-import { GobanCanvas } from "../GobanCanvas";
+import { GobanCanvas, GobanCanvasConfig } from "../GobanCanvas";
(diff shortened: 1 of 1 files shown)


anoek:
Very cool, thanks!

PR #103, opened 2022-12-08
onTap: remove unnecessary stone removal/scoring_mode code

benjaminpjones:
Now that this section is covered in tests, I figure a little refactoring is in order.

- cf600d4: add one more test to cover the ctrl-click case.  This passes before and after the refactor
- 04e40ee: Removes scoring_mode and ctrl/alt/metakey checks since onTap will never see these cases (see code snippet below)
- e1ef8d9: Remove the second socket message that gets sent when clicking a stone to remove it. 
    - I tested this manually and removing this seems to make no difference to the end user.  At the very least, it is unclear what it is supposed to mean.
    - Not only cleans up not network traffic, but also the code by reducing nesting.

----

The only place `onTap` is called is in `pointerUp()`.  This heavily simplified snippet should illustrate why `scoring_mode` and `ctrlKey`/`metaKey`/`altKey` are never true within `onTap()`:

```
const pointerUp = (ev: MouseEvent | TouchEvent, double_clicked: boolean): void => {
      ...
    if (this.scoring_mode) {
        // handle score estimate click
        return;
    }

    if (ev.ctrlKey || ev.metaKey || ev.altKey) {
        // add coordinates to chat
        return;
    }

    if (this.mode === "analyze" && this.analyze_tool === "draw") {
        // do nothing
    } else {
        this.onTap(ev, double_clicked, right_click);
    }
}
```

---

Screengrab of some manual testing.  localhost (with these changes) is left, beta (without these changes) is right. Note, when sending messages from beta, there are more socket messages, but the behavior in the UI is the same.

output

### DIFF
--- a/src/__tests__/GobanCanvas.test.ts
+++ b/src/__tests__/GobanCanvas.test.ts
@@ -5,6 +5,7 @@
 (global as any).CLIENT = true;
 
 import { GobanCanvas, GobanCanvasConfig } from "../GobanCanvas";
+import { GobanCore } from "../GobanCore";
 import { AUTOSCORE_TOLERANCE, AUTOSCORE_TRIALS } from "../GoEngine";
 import { GoMath } from "../GoMath";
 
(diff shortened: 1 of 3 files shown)


src/__tests__/GobanCanvas.test.ts
```diff
@@ -356,15 +374,6 @@ describe("onTap", () => {
                 stones: "babbbabbbbba",
```
benjaminpjones, line 374: I noticed there is a lot of redundancy in this message.  I wonder what's going on in `toggleMetaGroupRemoval()` that is causing this...

anoek:
Thanks for the cleanup!

PR #104, opened 2022-12-11
Oje sequence to moves

GreenAsJade:
Adds a function to get an `Array<JGOFMove>` from an OJE position (which is a move-sequence-string)

This is handy if you need to initailize a Goban from an OJE position :)

I put this into the `GoMath` file, because it seems to be `GoMath`.  I discovered I don't really know why tons of functions in there are inside the GoMath class.

I could't use my new function in the way that I wanted to do when it was in the class.

### DIFF
--- a/src/GoMath.ts
+++ b/src/GoMath.ts
@@ -37,6 +37,24 @@ export interface BoardState {
 
 type BoardTransform = (x: number, y: number) => { x: number; y: number };
 
+// OJE Sequence format is '.root.K10.Q1'  ...
+export function ojeSequenceToMoves(sequence: string): Array<JGOFMove> {
+    const plays = sequence.split(".");
+
+    if (plays.shift() !== "" || plays.shift() !== "root") {
+        throw new Error("Sequence passed to sequenceToMoves does not start with .root");
+    }
+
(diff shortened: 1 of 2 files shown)


src/GoMath.ts
```diff
@@ -37,6 +37,24 @@ export interface BoardState {
...
+    const plays = sequence.split(".");
+
+    if (plays.shift() !== "" || plays.shift() !== "root") {
+        throw new Error("Sequence passed to sequenceToMoves does not start with .root");
+    }
+
+    const moves = plays.map((play) => {
+        if (play === "pass") {
+            return { x: -1, y: -1 };
+        }
```
benjaminpjones, line 51: Nit: `GoMath.decodeGTPCoordinate()` knows how to handle `pass`, so it's not necessary to check here.

https://github.com/online-go/goban/blob/3fd86d45a63170a1641dec96a2c6313c868ac6be/src/__tests__/GoMath.test.ts#L108-L111
https://github.com/online-go/goban/blob/3fd86d45a63170a1641dec96a2c6313c868ac6be/src/GoMath.ts#L221-L223

anoek:
Thanks for the patch!

PR #105, opened 2022-12-12
Remove foreachNeighbor() as an argument to GoMath

benjaminpjones:
`GoMath` appears not to use this function, but rather implements [its own flood fill](https://github.com/online-go/goban/blob/a36c53c93ffdbc1f1d8452ba36e4ba68e87140eb/src/GoMath.ts#L53-L77).

Also, since the comment `/** Public for usage in GoStoneGroup */` is not accurate, I made these methods private.


anoek:
Yep, thanks!

PR #107, opened 2022-12-31
Use EventEmitter3 types

benjaminpjones:
## Background

For testing, it would be helpful to have `goban.eventNames()` and `goban.listenerCount()`.  I looked at adding those methods similar to the existing ones, but I thought it might be good just to pull the interface from EventEmitter3 directly.  It looks like [they've had a d.ts file for many years now](https://github.com/primus/eventemitter3/issues/67#issuecomment-250670758).

## Notes

The type interface for the generic arg is a bit different - the values are function signatures instead of a single data object.  This is more flexible, supporting multiple arguments, but OGS doesn't really make use of that, so I made a type transformer `LegacyEventsShim` to convert our existing event interfaces to something EventEmitter3 can consume.

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -28,7 +28,8 @@ import {
 } from "./JGOF";
 import { AdHocPackedMove } from "./AdHocFormat";
 import { _ } from "./translate";
-import { TypedEventEmitter } from "./TypedEventEmitter";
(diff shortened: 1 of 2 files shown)


anoek:
Very cool, yeah we should just ditch the TypedEventEmitter entirely then I'd say. Nice to have the shim so we don't have to do that all in one go though.

PR #111, opened 2023-02-16
[cleanup] Remove strict_seki_mode

benjaminpjones:
Strict seki mode isn't used on the site (https://github.com/online-go/online-go.com/pull/1861), but adds complexity to code, including Goban's scoring code.  Let's remove it?

## Testing

- I ran `yarn build` on the ogs-node package, and TS doesn't complain.
    - I haven't had luck running the node server though, so would be worth additional testing on beta.
    - I would want to check that GameData.strict_seki_mode gets set to `false`, not `undefined` for backwards compatibility.
- I ran `yarn dev` on the React package, and played a game.  The server is resilient to the `strict_seki_mode` missing from "removed_stones/accept" message.

## Notes on mobile

- Android sends `strict_seki_mode: false` in the `removed_stones/accept` message.
- iOS does not send `strict_seki_mode` in the `removed_stones/accept` message.
- I believe both still expect `strict_seki_mode: boolean` in the GameData message, but hopefully this PR does not remove that field

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -86,9 +86,6 @@ export interface GoEnginePlayerEntry {
     /** The accepted stones for the stone removal phase that the player has accepted */
     accepted_stones?: string;
 
-    /** Whether or not the player has accepted scoring with strict seki mode on or not */
-    accepted_strict_seki_mode?: boolean;
-
     /** XXX: The server is using these, the client may or may not be, we need to normalize this */
     name?: string;
     pro?: boolean;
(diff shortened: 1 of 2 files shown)


anoek:
I'm slightly reluctant to remove this, it's technically needed for proper Japanese scoring. While the expectation from online play is that we don't enforce strict seki mode, the pedantic part of me still wants to be able to support it. I'm also not entirely sure what removing this would do to the display of old games that had this flag set, they might look fine because we store the dead stones but we might not compute the correct score for presentation, I'm not entirely sure.

benjaminpjones:
Ah fair enough, thanks for your consideration anyways!

> I'm also not entirely sure what removing this would do to the display of old games that had this flag set, they might look fine because we store the dead stones but we might not compute the correct score for presentation, I'm not entirely sure.

Yeah I think that's accurate - from my understanding of how this is connected, it would mark territory-in-seki as points if it wasn't explicitly marked "removed".  I didn't consider this a big concern, since this option has been hidden for so long, but I can see how it would be a poor experience to look at an older game with an outcome-score mismatch.

PR #114, opened 2023-02-21
Remove GoStoneGroup.dame (unused)

benjaminpjones:
I guess this is related to the `dame` parameter used in `floodFill()`, but it doesn't seem to be used in `GoStoneGroup`.

I searched around both of the downstream codebases.  Obviously there are mentions of *dame*, but I'm pretty sure they didn't involve GoStoneGroup.

Tested: `yarn test` `yarn build-debug`


benjaminpjones:
It looks like there are many more unused fields in this class.  I am going to make a single PR with all the removals to prevent merge conflicts.

PR #115, opened 2023-02-21
Remove unused fields in GoStoneGroup.

benjaminpjones:
Supercedes  #113 and #114

The internals of `GoStoneGroup` do not use the following fields:

- `probable_color`
- `dame`
- `is_probably_dead`
- `is_probably_dame`
- `adjacent_white`
- `adjacent_black`

Because the fields are public, it's a little difficult to verify they aren't used, but `GoStoneGroup` is only constructed in `GoStoneGroups` and `GoStoneGroups` is only constructed in two places ([GoEngine](https://github.com/online-go/goban/blob/16481d6b1b678766d18c248c66f689e28a548e7e/src/GoEngine.ts#L1810), [ScoreEstimator](https://github.com/online-go/goban/blob/b1313e44a143e5903f58699614d231ada6800d71/src/ScoreEstimator.ts#L830)). Both scopes are short and the type checker should catch any erroneously removed fields.

Tested: `yarn build-debug`, `yarn test`

### DIFF
--- a/src/GoStoneGroup.ts
+++ b/src/GoStoneGroup.ts
@@ -27,20 +27,14 @@ export interface BoardState {
 }
 
 export class GoStoneGroup {
-    probable_color: JGOFNumericPlayerColor;
-    dame: boolean;
     corner_groups: { [y: string]: { [x: string]: GoStoneGroup } };
     points: Array<Intersection>;
     neighbors: Array<GoStoneGroup>;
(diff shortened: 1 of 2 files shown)


anoek:
Thanks for the cleanup, I think I used to use those fields for scoring stuff a long long time ago, but yeah long since abandoned I think.

PR #116, opened 2023-03-13
Typechecked GobanSocket for talking to the OGS servers, and protocol documentation

anoek:
This PR introduces the new "GobanSocket" helper class to connect to the OGS servers.

In addition to having the protocol and responses type checked, this will also serve as the official documented protocol specification for third party clients connecting to OGS.

- [X] Core websocket implementation
- [X] Client to server protocol types and documentation
- [x] Server to client protocol types and documentation
- [x] Optional promise message responses
- [x] Generated documentation hosted somewhere
- [x] Tests


src/GobanSocket.ts
```diff
@@ -115,22 +123,36 @@ export class OGSSocket extends EventEmitter<Events> {
...
 
+    public authenticate(authentication: DataArgument<ClientToServer["authenticate"]>): void {
+        this.authentication = authentication;
+        this.send("authenticate" as keyof SendProtocol, authentication as any);
+    }
+
     private sendAuthentication(): void {
-        //console.log("TODO: Implement authentication");
+        if (this.authentication) {
+            this.send("authenticate" as keyof SendProtocol, this.authentication as any);
```
benjaminpjones, line 133: I think we can get rid of some type assertions by telling TS that `SendProtocol` extends `{"authenticate": ..., "net/ping": ...}`.  I'll open a pull request trying this out!

benjaminpjones (commented):
Finally got a chance to look at these changes - for the documentation alone, this PR is gold!  😁

src/__tests__/GobanCanvas.test.ts
```diff
@@ -348,14 +367,14 @@ describe("onTap", () => {
         );
 
         // Unmodified clicks in stone removal send a "game/removed_stones/set" message
-        expect(mock_socket.send).toBeCalledTimes(0);
+        await sleep(50);
```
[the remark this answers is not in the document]
anoek: Yeah I think I arrived there in a round about way, at some point I was having trouble with the fake timer stuff. I just tried swapping this out back to the fake timer stuff though and it worked just fine so I think it was just a relic of trying to workaround something bad I was doing at some point. 71f3229f3eb94f2e9fafb878b5affef7473ebcd6 fixes it, thanks for the catch.

src/GobanSocket.ts
anoek: Yeah good call, and thanks for the PR :)

PR #118, opened 2023-03-20
Add base class for ClientToServer events.

benjaminpjones:
See https://github.com/online-go/goban/pull/116#discussion_r1141649945

This should help with the type hints in GobanSocket where the "authenticate" and "net/ping" messages are expected.

### DIFF
--- a/src/GobanSocket.ts
+++ b/src/GobanSocket.ts
@@ -16,7 +16,7 @@
 
 import { EventEmitter } from "eventemitter3";
 import { niceInterval } from "./GoUtil";
-import { ClientToServer, ServerToClient } from "./protocol";
+import { ClientToServer, ClientToServerBase, ServerToClient } from "./protocol";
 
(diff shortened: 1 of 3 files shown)


src/GobanSocket.ts
```diff
@@ -145,14 +145,11 @@ export class GobanSocket<
...
-                        client: Date.now(),
-                        drift: this.clock_drift,
-                        latency: this.latency,
-                    } as any,
-                );
+                this.send("net/ping", {
+                    client: Date.now(),
+                    drift: this.clock_drift,
+                    latency: this.latency,
+                } as DataArgument<SendProtocol["net/ping"]>);
```
benjaminpjones, line 152: Unfortunately still a type assertion here.  *I think* the issue is that TypeScript has no good way of knowing DataArgument<T> is contravariant, and so doesn't realize this send call is legit for classes that extend `ClientToServer`.

anoek:
Perfect thanks :)

Issue #121, opened 2023-06-14
Active chat input box causes auto scroll to it when playing a move on a touchscreen


andymarden:
see https://github.com/online-go/online-go.com/issues/2270

PR #122, opened 2023-06-14
Issue 121 - blur active input element on goban touch event

andymarden:
fixes https://github.com/online-go/goban/issues/121


anoek:
Thanks for finding and fixing that!

PR #130, opened 2023-10-02
Ping timeout event

GreenAsJade:
Gives GobanSocket a ping timeout event.


GreenAsJade:
This is on top of #129

PR #137, opened 2023-11-08
Score estimator more cleanup

benjaminpjones:
A few more cleanup opportunities that became apparent after spending more time with the ScoreEstimator code.


benjaminpjones:
I removed the commit to get rid of `ScoreEstimator.territory`.  I still believe it can be removed (see fbbb8ed172f92a97dc67db5781a3592654e371b3), but since the consuming code is in `GobanCanvas`, I would like to add more test coverage before removing it.

PR #138, opened 2023-11-11
Add ScoreEstimator tests and raise threshold

benjaminpjones:
Coverage for the file is 91% and global coverage is now 60%.

The remaining 10% is mostly some unused functions in SEGroup which I think we can get rid of (#139).


anoek:
Very nice :)

PR #139, opened 2023-11-11
Deprecate SEGroup

benjaminpjones:
With some minor additions, `GoStoneGroup` can completely replace the `SEGroup` in the score estimator.


anoek:
Awesome :)

PR #142, opened 2023-12-19
Fix ScoreEstimator bug for Japanese scoring

benjaminpjones:
This bug was introduced in d6194f4c46fdf90b8a12de091b038939230b8ca1 and caused captures to be multiplied by a factor of `height`.

https://forums.online-go.com/t/in-game-score-estimator-bugged/50046

---

Testing: I added a unit test based on [stone_defender's demo board](https://online-go.com/review/1190037).  8x8 instead of 20x20.

### DIFF
--- a/src/ScoreEstimator.ts
+++ b/src/ScoreEstimator.ts
@@ -531,12 +531,12 @@ export function adjust_estimate(
                 }
             }
         }
+    }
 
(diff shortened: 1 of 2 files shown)


anoek:
Thanks for the fix!

PR #144, opened 2024-01-12
Add the area handicap scoring bonus in Ing rules

dexonsmith:
Based on <https://www.cs.cmu.edu/~wjh/go/rules/KSS.html>, white should get an N-point handicap bonus in handicap games.  This is not from the official documentation of the Ing rules, but was requested by the AGA president.


dexonsmith:
Not 100% sure we should land this? But I think it makes sense.

dexonsmith:
I'm going to ask in the forums.

dexonsmith:
> I'm going to ask in the forums.

https://forums.online-go.com/t/should-ing-rules-have-give-a-handicap-bonus-to-white

anoek:
This seems like the right thing to do

PR #145, opened 2024-01-14
Support for supplying a url for stone images via Goban hook

GreenAsJade:
In support of https://github.com/online-go/online-go.com/issues/2481


anoek:
Converting to draft since this would introduce another stone before we're ready for it, pending https://github.com/online-go/online-go.com/pull/2498

GreenAsJade:
Note: Custom board is pretty much identical to Plain board, somewhat on purpose.   The UI is kinda interesting in that for custom board, both the colour selectors and the URL can apply at once - seems to be useful, having this second theme made it easy to lay out tidily.

PR #146, opened 2024-01-19
Set 'handicap' from GoEngine.computeScore to the handicap bonus

dexonsmith:
This commit changes `GoEngine.computeScore()` to return `0` for `handicap` when it does not contribute to the score.  Indirectly, this changes the scoring summary in the web-client to not show the handicap in this case.

This fixes an inconsistency in the API/return, which forwarded to the web-client and caused user confusion.

Previously it was neither consistently equal to the handicap (see AGA) nor the handicap scoring bonus (see Japanese/Korean/New Zealand).

- For Chinese/Ing, it was both the handicap scoring bonus and the handicap, which happen to be equal.
- For AGA, it was the handicap scoring bonus, which is `handicap-1`.
- For Japanese/Korean/New Zealand, it was the handicap (scoring bonus is 0).

Now, for all rulesets, it's the handicap scoring bonus (if any).  For a client computing the score, this is what's relevant.  This is similar to how `stones` is returned for rulesets where it's relevant to the score.

It might be nice for the web-client to show the number of handicap stones somewhere in the main view, but (a) it wasn't doing that previously and (b) `computeScore()` isn't the right way to get that info.

Fixes https://github.com/online-go/online-go.com/issues/1469

### DIFF
--- a/src/GoEngine.ts
+++ b/src/GoEngine.ts
@@ -1757,7 +1757,7 @@ export class GoEngine extends EventEmitter<Events> {
                 territory: 0,
                 prisoners: 0,
                 scoring_positions: "",
-                handicap: this.handicap,
+                handicap: this.getHandicapPointAdjustmentForWhite(),
                 komi: this.komi,
             },
             black: {
@@ -1770,10 +1770,6 @@ export class GoEngine extends EventEmitter<Events> {
                 komi: 0,
             },
         };
-
-        if (this.aga_handicap_scoring && ret.white.handicap > 0) {
-            ret.white.handicap -= 1;
-        }
 
         let removed_black = 0;
         let removed_white = 0;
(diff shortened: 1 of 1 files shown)


dexonsmith:
In the web-client, we now get the following for a 3-stone game after handicap stones have been placed and 1 move by white:

Chinese (no change):
chinese

AGA (no change):
aga

Japanese (was 3, now skipped because 0):
japanese

New Zealand (was 3, now skipped because 0):
nz

dexonsmith:
Oops; forgot to `commit --amend` after fixing a bug; force pushed to 7019ead29812e9ddc46e4e4bf0f6d3afeff0ed49, which should get the linter passing.

pdg137:
This sounds great! Maybe the English on that line should say "Handicap bonus" to reduce confusion.

dexonsmith:
> This sounds great! Maybe the English on that line should say "Handicap bonus" to reduce confusion.

Maybe... seems like it'd be better to consider separately, since the only user-visible changes in this commit are to hide the line when it _doesn't_ apply. (In any case, that'd be a change in the web-client.)

anoek:
Hmm, if I'm recalling correctly we had that in there for rules that it didn't matter because some vocal users wanted to quickly see the handicap stones given in the game.

dexonsmith:
> Hmm, if I'm recalling correctly we had that in there for rules that it didn't matter because some vocal users wanted to quickly see the handicap stones given in the game.

Well, it's the wrong number for AGA, so it's not serving that purpose well.  Maybe we can add the number of handicap stones somewhere?

dexonsmith:
> > Hmm, if I'm recalling correctly we had that in there for rules that it didn't matter because some vocal users wanted to quickly see the handicap stones given in the game.
> 
> Well, it's the wrong number for AGA, so it's not serving that purpose well. Maybe we can add the number of handicap stones somewhere?

See https://github.com/online-go/online-go.com/pull/2507.

anoek:
Thanks, good work @dexonsmith

PR #148, opened 2024-01-31
Add an interface declaration for GobanCanvas

GreenAsJade:
… so it's easier to see what it can do.

I'm not married to this idea, but it seems like it should be helpful?


anoek:
I like it, I have it in my mind to experiment with switching to SVG based boards sometime in the future too so a common interface makes a lot of sense for that eventuality.

PR #149, opened 2024-01-31
Change terminology from "transparency" to "opacity" 

GreenAsJade:
… when it's opacity we're dealing with.

We don't have to do this, but https://github.com/online-go/online-go.com/pull/2533 needs it.


anoek:
Yep, thanks!

PR #158, opened 2024-05-02
Fix jest

benjaminpjones:
As suggested by [this StackOverflow answer](https://stackoverflow.com/a/77810578/5001502), I deleted yarn.lock and rebuilt it from scratch:

```
rm yarn.lock
yarn install
```

Tests work on my machine now, we shall see if CI feels the same.

Update: CI is a success!

Screenshot 2024-05-01 at 8 31 11 PM


anoek:
Huh, cool, thanks for the fix!

PR #161, opened 2024-05-31
Autoscore

anoek:
This patch includes an improved autoscore system


benjaminpjones:
cc: @pdg137

Looks like [V3](https://pdg137.github.io/autoscore/v3.html)..  Or at least something like it? 😃

PR #167, opened 2024-06-25
Fix values of paths for clamshell lines

lykahb:
The mutable variable for the clamshell lines isn't reset for each line. I couldn't check the fix locally - the page fails to load test.js after running `yarn run dev`.

The problem:
image


anoek:
Good catch, thanks for finding and fixing that!

I'll get the dev env back in order, I must have broken during some recent refactoring.

PR #174, opened 2024-08-31
added "none" as a possible return value of getCoordinateDisplaySystem

andywng:
Summary of change: "none" is being added as a possibly return value of getCoordinateDisplaySystem.

Why: This change is required for https://github.com/online-go/online-go.com/pull/2804 to work properly.


anoek:
Closing on account of the None option already existing in the board label position

PR #177, opened 2024-09-14
Support a scale parameter for the font size for stone markings.

GreenAsJade:


anoek:
Hmm something with this messed up the top/sub positioning:

image



Should look more like this:

image

GreenAsJade:
Screenshot 2024-09-17 at 11 00 03 AM (2)

Note: I found another parameter `sub_triangle`, which doesn't seem to be tested in this board.   It's such a teeny tweak in the maths, I'm kinda hoping it just does the right thing if/when needed  😝

If we know the "right context" for this to be tested in (how to actually set it in main.tsx testcase), I'm happy to add it: it's a parameter of the `MoveTree` interface, and it's also a valid `MARK_TYPE".

GreenAsJade:
Oh wait, I just saw that for some reason the original "sub" is actually centered in the square.  Is that intended? 

I imagined a sub would be positioned in a subscript position, so I didn't notice that I'd "broken" that!

GreenAsJade:
Latest (note that my Mac gives me bold font on stone lettering!)

Screenshot 2024-09-17 at 11 39 08 AM (2)

anoek:
Awesome, thanks!

PR #185, opened 2025-04-23
Improve stone shadows

kevincoleman:


anoek:
The unrelated line changes look a bit wonky on the Play page:

image

And learning page

image

what was the use case you were addressing with that change?

kevincoleman:
Hmm. Great point. I’ll see if I can figure out how it scales a bit better. I’m more used to working in CSS for this kind of edit, but there’s probably a solution in SVG too.

kevincoleman:
@anoek Not sure why, but GitHub dropped the link between this PR and my branch. Please see https://github.com/online-go/goban/pull/194.

I was trying to thicken the lines a bit so the shadows themselves would interact better with the board, but in retrospect that wasn’t really necessary. This PR has another commit to roll that back.

PR #203, opened 2025-09-24
Update internal links to use @link syntax [Fixes #201]

### LINKED ISSUE #201: docs: [[DoubleBrackets]] aren't linking
Example:

https://github.com/online-go/goban/blob/22ebe12ebb7c18da349179b72216a2eaa866444a/src/engine/protocol/index.ts#L30

It looks like these are meant to turn into links.  But that's not happening:

Image

Raised here: https://forums.online-go.com/t/rt-api-game-data-documentation/58099/5

lazerwalker:
https://docs.online-go.com/goban/modules/protocol.html contains the text `[[ClientToServer]]` and `[[ServerToClient]]`, which are presumably intended to be hyperlinks to the relevant docs pages. However, that syntax is no longer supported in typedoc 0.25, which this project is currently using, so instead of links we have plaintext in awkward brackets.

See https://typedoc.org/documents/Changelog.html#breaking-changes-4:
> Dropped support for legacy [[link]]s, removed deprecated Reflection.findReflectionByName.

This migrates those two links to the modern `{@link ClientToServer}` syntax. A brief projectwide search for `[[` suggests these two instances are the only two `[[links]]` in the docs.

Tested changes locally with `yarn typedoc`. Test suite and prettier are also clean, although that's naturally expected for a two-word docs change.

Cheers!

### DIFF
--- a/src/engine/protocol/index.ts
+++ b/src/engine/protocol/index.ts
@@ -27,7 +27,7 @@
  *
  *    `[command: string, data: any, id?: number]`
  *
- * where command is one of the commands listed in the [[ClientToServer]] protocol,
+ * where command is one of the commands listed in the {@link ClientToServer} protocol,
  * data is the data associated with that command, and id is an optional
  * request id that will be echoed in a response if a response is requested.
  *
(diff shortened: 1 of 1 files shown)


anoek:
Thanks!

PR #206, opened 2025-10-05
Modify game phase to include 'stone removal'

benjaminpjones:
More than half of acceptance events will be in the phase "stone removal"


anoek:
Thanks!

PR #236, opened 2026-04-16
A few type updates to match the actual wire format

lazerwalker:
Two small things!

- In practice, GobanEngineConfig objects contain a `private` flag that isn't reflected in `GobanEngineConfig`. I don't have the cleanest sense of the divide between `GobanEngineConfig` and `BoardConfig`, but it should definitely exist on one of them, and my gut instinct was this belongs alongside e.g. `ranked`
- The challenge payload structure is both missing `komi_auto`, and has a bunch of fields that are currently just typed as `string` where the actual possible input is narrower and known (and in the case of challenger_color, we have a dedicated type, but it's missing an option the client sends today!)

Should be no practical runtime changes, just trying to help other folks by getting the types to match the actual wire format whenever I notice discrepancies.

Cheers!

### DIFF
--- a/src/engine/GobanEngine.ts
+++ b/src/engine/GobanEngine.ts
@@ -134,6 +134,7 @@ export interface GobanEngineConfig extends BoardConfig {
     moves?: GobanMovesArray;
     move_tree?: MoveTreeJson;
     ranked?: boolean;
+    private?: boolean;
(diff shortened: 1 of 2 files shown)


anoek:
Awesome, thanks!

PR #240, opened 2026-04-18
Update JGOFAIReview id

lazerwalker:
Yet another teeny tiny "the wire types don't match the TS types" fix.

### DIFF
--- a/src/engine/ai/AIReviewData.ts
+++ b/src/engine/ai/AIReviewData.ts
@@ -105,7 +105,7 @@ export class AIReviewData extends EventEmitter<AIReviewDataEvents> implements JG
         this.emit("destroy");
     }
 
-    public get id(): string {
+    public get id(): number {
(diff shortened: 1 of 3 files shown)


lazerwalker:
Fun, so my local build setup was apparently wonky, I swear I did explicitly run `npm run test` locally and it came up clean, but at least I'm able to get the valid failures now. Checks and test suite both clean, at least locally.

anoek:
Thanks!

lazerwalker:
FYI, I'm going to open a new PR in a little bit (there are a few other inaccuracies with JGOFAIReview and its ilk), but, frustratingly, this should actually be `string|number`. The `/game/:id/ai_reviews` API v1 endpoint returns numeric ids, while the WS connection returns string ids. Presumably not the highest-priority, since JS/TS plays strings vs numbers fast and loose, but just mentioning for now.

PR #241, opened 2026-04-19
Goban/baduktv f59810

RainWaterProductions:
This PR is for https://github.com/online-go/online-go.com/issues/3509

It will require a second PR in the main repo as well. They should be merged together.

The BadukTV stones still need fixing, I'll do that today.


anoek:
The images here differ from the ones in the main repo and these ones cause alignment issues for the stones. Same open questions about ownership and permission of these images as well.

RainWaterProductions:
@anoek 
**Yes I wasn't done with the stones.** I did say that in the initial comment of this PR but I guess I forgot to set it to Draft instead of Open.
Same with the images differing. I'll fix both of these things today.

**Regarding the ownership and permissions:**
- The BadukTV images are from the upsided's repo which I linked above. [The repo uses the MIT license](https://github.com/upsided/Upsided-Sabaki-Themes?tab=MIT-1-ov-file#readme). That should work, right?
- The "Community Favorite" image stems from the [OGS forum here](https://forums.online-go.com/t/board-backgrounds-library/29388/26), from the user taatelikakku who says he applied a "color ramp" to a picture from [Inigo Quilez](https://iquilezles.org/). I'm not sure exactly if applying a color ramp modifies the ownership of the picture, I'd assume it does. If this is too vague, then we could use the favorite community background where we have the permission for sure.

This would be second place for Community Favorite and it completely belongs to the OGS user Feijoa who gave us permission in chat: https://forums.online-go.com/t/board-backgrounds-library/29388/217

PR #242, opened 2026-04-20
Add support for opt-in fuzzy stone placement

Seth-Rothschild:
This PR adds support for fuzzy stone placement. The Sabaki implementation is pretty complex. It adjusts stones post-hoc if they're too close to each other. I think that's likely more effort than you would want to maintain here. So instead for this implementation:

1. We have a user preference for `getFuzzyPlacementEnabled` which should default to false.
2. We decrease the stone size to 90% when the setting is enabled
3. We generate a (not actually very random, see below) offset deterministically based on seed and stone position. Scale it to 5% of the square size so stones can't overlap even in the worst case scenario.
4. Finally add `transform` to the svg in src/Goban/SVGRenderer.ts with that offset. This moves the stones and annotations to stay on the stones. It also might move analysis colors and the star points if they're below the stones according to Claude? Everything looked ok to me on visual inspection.

This seems to work pretty nicely and should not impact users unless they explicitly opt into this. That being said, I'd appreciate another set of eyes since these files are big and I know even less about them than I know about the main OGS code.

*Note on the "prng":*
Definitely not actually random. We take gameid mod 1000 / 1000 (some fixed number between 0 and 1), a stone index (fixed integer between 0 and number of squares-1), and whichever irrational number we want (I used sqrt(2) and sqrt(3). Then truncate to [0. 1) and move values to (-1., 1). There's **absolutely no reason** to use this function or this function signature instead of another function. Was just looking for something deterministic and quick to compute. 

example image on demo board
image

example image for the liveplay games
image

### DIFF
--- a/src/Goban/Goban.ts
+++ b/src/Goban/Goban.ts
@@ -96,6 +96,19 @@ export abstract class Goban extends OGSConnectivity {
     private _evaluation: number = 0.5;
     private _evaluation_bar_width: number = 8;
 
+    private prng(seed: number, irrational: number): number {
+        return (((seed + irrational) % 1) - 0.5) * 2;
+    }
+
+    public getStonePlacementOffset(i: number, j: number): { x: number; y: number } {
+        const max_shift = 0.05 * this.square_size;
+        const seed = ((this.game_id || 0) % 1000) / 1000;
+        const stone_index = i * this.width + j;
+        const dx = this.prng(seed, stone_index * Math.sqrt(2)) * max_shift;
+        const dy = this.prng(seed, stone_index * Math.sqrt(3)) * max_shift;
+        return { x: dx, y: dy };
+    }
+
     constructor(config: GobanConfig, preloaded_data?: GobanConfig) {
         super(config, preloaded_data);
 
--- a/src/Goban/Goban.ts
+++ b/src/Goban/Goban.ts
@@ -103,7 +103,7 @@ export abstract class Goban extends OGSConnectivity {
     public getStonePlacementOffset(i: number, j: number): { x: number; y: number } {
         const max_shift = 0.05 * this.square_size;
         const seed = ((this.game_id || 0) % 1000) / 1000;
-        const stone_index = i * this.width + j;
+        const stone_index = 1 + i * this.width + j;
         const dx = this.prng(seed, stone_index * Math.sqrt(2)) * max_shift;
         const dy = this.prng(seed, stone_index * Math.sqrt(3)) * max_shift;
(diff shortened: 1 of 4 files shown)


Seth-Rothschild:
I'm not going to specifically fix for (0, 0) per the review: though if there's a different rng function we'd like here in general I'd be happy to implement. I can live with 0,0 always living along the y=x line especially for offsets this small. It's hard to see an individual stone offset at all let alone notice that one is on a particular line.

I think canvasrenderer is on its way out? I'm comfortable not implementing for that but it seems like it wouldn't add a ton of complexity to add it in.

anoek:
Neat, I'll review this more tomorrow but wanted to say post placement stone adjustments would be fine to maintain if you felt inclined to implement those.

Definitely don't need a canvas implementation, it'd be way more trouble than it's worth getting the rendering right.

For the top left 0,0, you could just add one to the coordinates and solve that so it's not always in the same place?

Seth-Rothschild:
> Neat, I'll review this more tomorrow but wanted to say post placement stone adjustments would be fine to maintain if you felt inclined to implement those.

I appreciate it!

For me (and my guess is for most people) just the little bit of offset is enough to scratch the itch. When you look closely at the Sabaki ones you can sometimes see the stone moving out of the way, which is fun but I think most people don't notice. Since I suspect it would have low impact, and since this should be much easier to read/debug I would vote for sticking with something similar to this for now.

> For the top left 0,0, you could just add one to the coordinates and solve that so it's not always in the same place?

That would absolutely be sufficient and would read better than messing with seed. I sort of feel like the signature should be something like

```
def prng((i, j), axis, seed):
    # something less convoluted like cosine(i+j+axis+seed)
    return (something in (-1, 1)).
```
the problem is just accidentally making correlations that are visually recognizable. Which this could be since i, j, and axis are less than the period of cosine...

anoek:
This looks great. I personally think the 0.45 scale is a bit excessive, I was playing around with `0.49` and it seemed to work well given the offsets aren't that big. Do you think we should keep it at 0.45 or do you want to adjust that before merge?

Do you want to address the 0,0 thing before merge, or are you happy with things? Visually it's hard to notice, but seems like an easy thing to fix too.

Seth-Rothschild:
I'm happy to adjust both. I don't know mathematically why .49 doesn't overlap more often, but I filled up another demo board and it really seems like it doesn't (at least at seed=0). Interestingly, if we were to use cosine that _would_ hit the maximum shift more often since sin/cosine spend more time near 1.

PR #246, opened 2026-05-31
Fix double-click-to-move after an opponent's pass (#3364)

### LINKED ISSUE #3364: `Double-click to move` does not work when the previous move is an opponent's pass
**Describe the bug**
The `Double-click to move` mode does not work when the previous move is an opponent's pass, acting like the `Submit-move button` mode.
After submitting the move with the button, the `Double-click to move` mode works correctly again until the opponent passes again.

I suspect the `Submit-move button` mode behavior is because I set `Time Control` to `None` in the game challenge. I have included a screenshot of the game information in case any of it is relevant.

The `Double-click to move` mode bug might act differently for `Live games`; I did not check.

**To Reproduce**
1. Set the settings at [https://online-go.com/settings/game](https://online-go.com/settings/game) to match the screenshot.
2. Make a game with the same rules as the screenshot.
3. Have your opponent pass.
4. Try double-clicking a vertex to make a legal move.

**Expected behavior**
I expect double-clicking a vertex to cause a move at that vertex, if the move is legal.

**Screenshots**
Image
Image

**Desktop**
- OS: Linux 6.18.2-2-cachyos
- Browser: Google Chrome
- Version: 143.0.7499.169 (Official Build) (64-bit)

yuki2006:
Fixes online-go/online-go.com#3364

## Problem

With `Double-click to move` enabled, double-clicking a vertex fails to play the move when the **previous official move was the opponent's pass**. The board acts as if it were in submit-button mode (a provisional stone you must confirm), recovering only after the next move is submitted normally — then breaking again on the next pass.

## Root cause

Double-click-to-move was detected **only** via the browser's native `dblclick` event (`onDblClick` → `pointerUp(ev, true)`). Each click is committed on `mouseup`; placing the provisional stone mutates the DOM under the cursor (after a pass the embedding app's UI changes, the board resizes and is redrawn), and the browser then **fails to synthesize the `click`/`dblclick` events** — so the second tap only removes the provisional stone and the double-click is lost.

Event sequences captured on a live game with real (CDP) mouse input:

| previous official move | events delivered to the board | result |
| --- | --- | --- |
| normal stone | `mousedown, mouseup, mousedown, mouseup, click, dblclick` | move submitted ✅ |
| opponent **pass** | `mousedown, mouseup, mousedown, mouseup` (no `click`, no `dblclick`) | nothing submitted ❌ |

This is why goban's own jsdom unit tests don't catch it — they dispatch a synthetic `dblclick`, so they always exercise the `double_clicked = true` path. The bug is that the *browser* stops emitting that event after the pass-triggered DOM mutation, so it only shows up with real input in the full app.

## Fix

Stop relying solely on the native `dblclick`. In `pointerUp` (both `SVGRenderer` and `CanvasRenderer`), detect the double-tap from timing + position: if a second release lands on the same square within `DOUBLE_CLICK_TIMEOUT_MS` (500 ms) and `double_click_submit` is enabled, treat it as a double-click. A `synthesized_double_click` flag makes a subsequently-fired native `dblclick` a no-op so a move is never submitted twice. Gated on `double_click_submit`, so submit-button mode is unchanged.

## Testing

- New regression test `test/unit_tests/DoubleClickFallback.test.ts` (drives two releases **without** a native `dblclick`) + existing `GobanSVG.test.ts` pass (15 tests total).
- `tsc --noEmit` clean for `src/`.
- Manual, in a live OGS game (correspondence, time control "none", two players): real double-click after the opponent's pass now submits the move (even though the browser still emits no native `click`/`dblclick`); double-click after a normal move submits exactly once; with `double_click_submit` off it stays in submit-button mode.
- TODO before merge: manual check on a touch device and on the Canvas renderer (only SVG was exercised live; OGS uses SVG via shadow DOM).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Goban/CanvasRenderer.ts
+++ b/src/Goban/CanvasRenderer.ts
@@ -448,8 +448,12 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
                         if (synthesized_double_click) {
                             /* Already handled via the timing fallback above; a
                              * native `dblclick` also arrived, so ignore it to
-                             * avoid submitting/placing twice. */
+                             * avoid submitting/placing twice. Reset the timing
+                             * baseline so the first click of the next move can't
+                             * chain off this (now stale) timestamp and be
+                             * mistaken for a double-click. */
                             synthesized_double_click = false;
+                            last_pointer_up_timestamp = 0;
                             this.onMouseOut(ev);
                             return;
                         }
--- a/src/Goban/CanvasRenderer.ts
+++ b/src/Goban/CanvasRenderer.ts
@@ -71,6 +71,10 @@ interface ViewPortInterface {
 
 const HOT_PINK = "#ff69b4";
 
+/* Max time between two releases on the same square for our timing-based
+ * double-click fallback to treat them as a double-click (#3364). */
+const DOUBLE_CLICK_TIMEOUT_MS = 500;
+
 export interface GobanCanvasInterface {
     engine: GobanEngine;
     move_tree_container?: HTMLElement;
@@ -346,6 +350,16 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
 
         let last_click_square = this.xy2ij(0, 0);
         let pointer_down_timestamp = 0;
+        /* Fallback double-click detection. The browser does not always
+         * synthesize a native `dblclick` event: if the DOM under the cursor is
+         * mutated between the two clicks the click/dblclick events are dropped.
+         * This happens e.g. after an opponent's pass, where placing the
+         * provisional stone triggers a board redraw, which left
+         * `double-click-to-move` behaving like the submit-button mode (#3364).
+         * We therefore detect the double-tap ourselves from timing + position
+         * in `pointerUp` and ignore the native `dblclick` when we already did. */
+        let last_pointer_up_timestamp = 0;
+        let synthesized_double_click = false;
 
         const pointerUp = (ev: MouseEvent | TouchEvent, double_clicked: boolean): void => {
             const press_duration_ms = performance.now() - pointer_down_timestamp;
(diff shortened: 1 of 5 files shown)


yuki2006:
Thanks — both points were valid; addressed in d339d3b1:

1. **Stale `last_pointer_up_timestamp` after native-dblclick suppression.** The suppression branch now also resets `last_pointer_up_timestamp = 0` (SVG + Canvas), so the first click of the next move can't chain off the stale timestamp and be mistaken for a double-click (the blitz case you described).

2. **`commonConfig()` ignored its argument.** It now takes an optional `Partial<SVGRendererGobanConfig>` and spreads it, so `double_click_submit: false` is actually applied. Added `expect(goban.double_click_submit).toBe(false)` as a guard, and both tests now spy on `sendMove` (asserting it's called with `move: "ee"` in the on case and not called in the off case) so they test the intended path independently of the mocked transport.

All 17 tests pass (2 fallback + 15 GobanSVG).

yuki2006:
Good call — done in 4010833a. The duplicated state machine now lives on the shared `Goban` base class: the state (`last_pointer_up_timestamp` / `last_pointer_up_square` / `synthesized_double_click`) and `DOUBLE_CLICK_TIMEOUT_MS` moved there, behind a `protected resolveDoubleClick(pt, double_clicked): "single" | "double" | "ignore"` helper. Each renderer's `pointerUp` now calls it in one place, so the threshold and logic live in a single spot. No behavior change.

`tsc --noEmit` clean; 32 tests pass (DoubleClickFallback + GobanSVG + GobanCanvas).

yuki2006:
Both valid — thanks, especially catching the Canvas gap. Fixed in 17078e5b:

1. **CanvasRenderer couldn't fire the fallback.** Right — canvas drove placement from `click`, which the browser also drops in the #3364 scenario, so `resolveDoubleClick` ran only once (or not at all). Mirrored `SVGRenderer`: `pointerUp` is now handled on `mouseup` and `click` is a no-op. The regression test now runs against **both** renderers (`describe.each`), and it does fail on canvas without this change.

2. **Non-synthesized native `dblclick` left `last_pointer_up_timestamp` set.** `resolveDoubleClick` now resets the timestamp on that branch too (clicks farther apart than our 500 ms window but within the OS double-click time), closing the same asymmetry as the synthesized branch.

Also updated the GobanCanvas tests to drive modifier-key / stone-removal clicks via mousedown+mouseup (its `simulateMouseClick` now forwards modifier keys), matching the mouseup model — that surfaced a pre-existing fake-timer leak between two of those tests, now gone.

Full suite: 331 tests pass; `tsc --noEmit` clean.

yuki2006:
Both valid — fixed in fd5e0370.

1. **Right-click priming a double-click.** `resolveDoubleClick` now returns `"single"` for non-primary-button releases *without* recording their timing, so a right-click followed by a primary click on the same square within the window is no longer taken as a double-click. The native `dblclick` only ever fired for the primary button, so this matches the previous behavior. Added a regression test (both renderers): right-click then left-click on the same square does not submit.

2. **Firefox context menu regression in CanvasRenderer.** The new `mouseup` handler now only handles `ev.button === 0`, so right-clicks keep flowing through the `contextmenu` handler that `preventDefault()`s the native menu — `mouseup` no longer clears `dragging` before `contextmenu` fires.

Full suite: 333 tests pass; `tsc --noEmit` clean.

anoek:
Thanks for digging into that and getting a fix together @yuki2006 !

PR #247, opened 2026-06-01
Last-move accessibility crosshair

superkeil:
## Summary

Adds an opt-in, high-contrast **crosshair** (full horizontal + vertical lines) through the last played stone, drawn **under** the stones, to help low-vision players locate the last move. Disabled by default; driven by the consuming app.

This is the **goban half** of a two-repo change. The web client side (preference keys, `configure-goban` wiring, the Accessibility settings section + live preview, and suppression on thumbnails) lives in a companion `online-go.com` PR that bumps the submodule to this commit once merged.

## What's here

- **`getLastMoveCrosshair()` callback** (`{ enabled, color, thickness }`) — read at draw time, like the other appearance getters (no watcher).
- **Canvas** (`CanvasRenderer`): a dedicated under-stone canvas (inserted behind the transparent stone canvas, z-index between shadow and stone) draws each line as a **single stroke**, cleared and redrawn wholesale on each last-move change → continuous lines, no residual segments when navigating moves. Attached lazily (no cost unless enabled).
- **SVG** (`SVGRenderer`): a dedicated crosshair layer beneath the stone cells holding two `<line>`s, rebuilt on redraw / last-move change.
- Gated on the existing `dont_draw_last_move`, plus a new per-board **`dont_draw_last_move_crosshair`** flag so thumbnail/preview boards can opt out while keeping the normal last-move marker.
- Thickness is relative (fraction of square size) so it scales with board size.

## Tests

- Unit tests (Canvas + SVG): crosshair layer is present when enabled and a last move exists; absent when disabled, when `dont_draw_last_move` / `dont_draw_last_move_crosshair` is set, or when there's no last move.
- Full suite green; lint + prettier clean.

## Manual testing

Verified visually in the web client (both renderers): lines run the full board, pass under the stones, follow each move, and erase cleanly on navigation; color and thickness controls work.

### DIFF
--- a/src/Goban.css
+++ b/src/Goban.css
@@ -63,7 +63,7 @@
 
     .CrosshairLayer {
         position: absolute;
-        z-index: var(--z-goban-crosshair-layer, 15);
+        z-index: var(--z-goban-crosshair-layer);
         left: 0;
         right: 0;
     }
--- a/src/Goban.css
+++ b/src/Goban.css
@@ -16,7 +16,7 @@
 
 .Goban {
     position: relative;
-    background-color: #DCB35C;
+    background-color: #dcb35c;
     display: inline-block;
     box-shadow: var(--goban-shadow);
 
@@ -61,6 +61,13 @@
         right: 0;
     }
 
+    .CrosshairLayer {
+        position: absolute;
+        z-index: var(--z-goban-crosshair-layer, 15);
+        left: 0;
+        right: 0;
+    }
+
     .GobanMessage table {
         position: relative;
         width: 100%;
@@ -75,10 +82,10 @@
     .GobanMessage table td div {
         border-radius: 3px;
         box-shadow: 5px 5px 3px rgba(0, 0, 0, 0.26);
-        color: #FF5500;
+        color: #ff5500;
         background-color: #262626;
-        padding: .3em;
-        margin: .4em;
+        padding: 0.3em;
+        margin: 0.4em;
     }
 
     .hidden {
--- a/src/Goban/CanvasRenderer.ts
+++ b/src/Goban/CanvasRenderer.ts
@@ -2282,13 +2282,20 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
         }
 
         /* Clear last move */
+        // Tracks whether the crosshair was already refreshed in this draw, so the
+        // "draw last move" block below doesn't redraw it a second time for the
+        // same cur_move.
+        let crosshair_synced = false;
         if (this.last_move && this.engine && !this.last_move.is(this.engine.cur_move)) {
             const m = this.last_move;
             delete this.last_move;
             this.drawSquare(m.x, m.y);
             // The last-move crosshair lives on its own canvas; refresh it whenever
             // the last move changes (live moves / navigation are targeted draws).
+            // This also clears it when navigating to a position with no last-move
+            // stone, which the "draw last move" block below cannot.
             this.drawLastMoveCrosshair();
+            crosshair_synced = true;
         }
 
         /* Draw last move */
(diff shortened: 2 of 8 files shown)


superkeil:
Thanks for the review — addressed in 22513aa0.

**SVG lazy attach:** `updateLastMoveCrosshair` no longer touches the DOM when the crosshair isn't shown. The layer is created and `insertBefore`-d only inside the shown path (and merely cleared when an existing layer should hide), mirroring the Canvas renderer. Disabled SVG boards no longer get an empty `<g class="crosshair-layer">` or repeated `insertBefore` calls.

superkeil:
Both addressed in eb8b9863:

1. **SVG `insertBefore` every move** — now guarded: `if (this.crosshair_layer.nextSibling !== this.grid_layer) { ... }`. It still re-positions after a `force_clear` (when `grid_layer` is recreated) but no longer mutates the DOM on every move.
2. **`willReadFrequently` on a write-only canvas** — dropped for the crosshair context (both creation and resize), so it stays GPU-accelerated. It's the only purely write-only layer; the shadow/pen/board canvases keep the hint as they read pixels back.

superkeil:
All three addressed in ded0ddcd:

1. **z-index fallback above stones** — dropped the `, 15` fallback; `.CrosshairLayer` now uses `var(--z-goban-crosshair-layer)` with no fallback, matching every other layer. When the vars are unset (standalone), it falls back to `auto` like the others and DOM order (the crosshair canvas is inserted before the board) keeps it under the stones.
2. **SVG 0.5px offset** — `updateLastMoveCrosshair()` now uses `this.metrics.mid` instead of `Math.round(ss / 2)`, so the lines sit exactly on the stone centres (and match the Canvas crosshair). (Left `drawLines()` untouched — that's the existing grid path.)
3. **SVG destroy cleanup** — `destroy()` now `delete`s `crosshair_layer` and `grid_layer`, so a late `updateLastMoveCrosshair()` early-returns and the detached subtree can be GC'd, mirroring the Canvas `detachCrosshairLayer()`.

Re-validated: goban unit tests + the web client e2e (both renderers) pass.

superkeil:
**#1 (willReadFrequently)** — fixed in f01eb5c2. You're right that my previous `getContext("2d")` was a no-op: `createDeviceScaledCanvas` already locks the context in with `willReadFrequently: true` on its first `getContext` call. I've passed the same option as the shadow/pen/board layers for consistency and removed the misleading comment. (Truly bypassing it would mean reimplementing the device-scaled allocation just for this tiny write-only layer — not worth it; it's CPU-backed like every other goban canvas.)

**#2 (drawn twice per move)** — keeping both calls deliberately. The suggested fix (drop the 'clear old' call, keep only 'set new') has a gap: the 'set new' block is guarded by `this.engine.board[j][i]` (a stone at cur_move), so navigating to a position with no last-move stone (game start / after a pass / back to the root) would never fire it and the crosshair would be left stale. The 'clear old' call is what clears it in that case. The redundant second pass is two line strokes within a single synchronous draw (same frame, no visible flicker), so I'm favouring correctness over the micro-optimization here.

superkeil:
Addressed in 9c4d4cae. The crosshair update now runs **once per move**: a local flag in the draw path records whether the 'clear old last move' block already synced it, and the 'draw new last move' block only calls it when that flag is unset (i.e. the first move, where the clear block doesn't run).

I kept the call in the 'clear' block rather than removing it: the 'draw new' block is gated on `engine.board[j][i]` (a stone at cur_move), so it can't clear the crosshair when navigating to a position with no last-move stone (game start / after a pass / back to the root) — the clear block handles that. Net effect: one redraw per move in all cases, no stale crosshair on navigation.

Verified: goban unit tests + the web-client e2e (both renderers) pass.

superkeil:
Addressed in `ed9391ca`.

**2. Stale `crosshair_ctx` on null `getContext` during resize — fixed.** The resize block now has the missing `else`:

```ts
if (ctx) {
    this.crosshair_ctx = ctx;
} else {
    delete this.crosshair_ctx;
}
```

I drop the stale reference rather than `throw` like the core layers do: the crosshair is an optional accessibility overlay, so on context loss it should silently skip drawing (the `!this.crosshair_ctx` guard in `drawLastMoveCrosshair` handles that) instead of aborting the whole board redraw.

**1. `attachCrosshairLayer` `appendChild` fallback — documented, no behaviour change.** This path is a verbatim copy of the long-standing `attachShadowLayer` fallback (same `insertBefore`→`appendChild`), and correct under-stone stacking relies on the consumer defining `--z-goban-crosshair-layer` — which OGS does. The `--z-goban-*` variables intentionally have no in-library fallback (consistent with the other z-index vars, settled earlier in this PR), so the same caveat already applies to the shadow layer. I added a comment on the fallback making the z-index dependency explicit, mirroring the shadow layer's existing note.

Verified: `tsc --noEmit` clean, crosshair unit tests 10/10.

anoek:
Thanks!

PR #252, opened 2026-06-18
Add layered custom board grid backgrounds

lykahb:
## Summary

Adds renderer support for custom board background assets that include a baked-in grid. The motivation is to make boards that look better - the vector grid doesn't fit some organic look. Although I included very stylized examples, even the regular wooden board would look better when the lines are a part of the image.

The changes are made with Codex. I've reviewed them and closely steered the technical design.

This keeps the existing custom board background behavior intact, and adds an optional per-board-size baked-grid background path for 9x9, 13x13, and 19x19 boards. The baked-grid image is rendered as a separate layer above the default board background and vector grid, so broken image URLs gracefully degrade to the vector-grid board.

Related online-go.com PR: https://github.com/online-go/online-go.com/pull/3593

## Screenshots

From testing on OGS locally. The grid is painted lines on the lawn
image

Screenshot of another theme I made, where grid blends with the theme more. The grid is formed by the gaps between the tiles.
image

## Details

- Adds optional `custom-board-grid-backgrounds` theme data.
- Resolves baked-grid backgrounds only for supported square board sizes.
- Uses explicit asset metadata for the current large-margin/no-baked-label profile, leaving room for future small-margin or baked-label profiles.
- Updates both SVG and canvas renderers to layer baked-grid backgrounds without removing fallback rendering.
- Keeps coordinate labels as renderer overlays.
- Fixes canvas custom URL stones to load plain stones instead of blank canvases, including move-tree stones. When testing, I thought that my changes broke the canvas rendering so I fixed it. But actually that bug exists upstream too. To test it, set the stones assets to URL and set the legacy canvas renderer.

## Compatibility

This is backward-compatible for existing goban consumers:
- The new theme data is optional.
- Existing custom board URLs/colors continue to work.
- Missing, broken, or unsupported baked-grid backgrounds fall back to the default board background and vector grid.
- Existing theme implementations remain valid.

## Testing

- Tested on the desktop and mobile browsers
- Tested for graceful degradation when background assets that have broken URL's

### DIFF
--- a/src/Goban/CanvasRenderer.ts
+++ b/src/Goban/CanvasRenderer.ts
@@ -112,9 +112,9 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
     private message_timeout?: number;
     private shadow_layer?: HTMLCanvasElement;
     private shadow_ctx?: CanvasRenderingContext2D;
-    private grid_layer: HTMLCanvasElement;
-    private grid_ctx: CanvasRenderingContext2D;
-    private grid_background_layer: HTMLDivElement;
+    private grid_layer?: HTMLCanvasElement;
+    private grid_ctx?: CanvasRenderingContext2D;
+    private grid_background_layer?: HTMLDivElement;
     private handleShiftKey: (ev: KeyboardEvent) => void;
 
     private last_move_opacity: number = 1;
@@ -168,19 +168,6 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
         /* TODO: Need to reconcile the clock fields before we can get rid of this `any` cast */
         super(config, preloaded_data as any);
 
-        this.grid_layer = createDeviceScaledCanvas(10, 10);
-        this.grid_layer.setAttribute("id", "grid-canvas");
-        this.grid_layer.className = "GridLayer";
-        const grid_ctx = this.grid_layer.getContext("2d");
-        if (grid_ctx) {
-            this.grid_ctx = grid_ctx;
-        } else {
-            throw new Error(`Failed to obtain drawing context for board grid`);
-        }
-
-        this.grid_background_layer = document.createElement("div");
-        this.grid_background_layer.className = "GridBackgroundLayer";
-
         this.board = createDeviceScaledCanvas(10, 10);
         this.board.setAttribute("id", "board-canvas");
         this.board.className = "StoneLayer";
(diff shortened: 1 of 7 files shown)


lykahb:
I think that Claude is right about the CSS escaping. However, I decided to revert the addition of `cssUrl` [44cc9cf](https://github.com/online-go/goban/pull/252/commits/44cc9cfc611b793fea98788dc47ebc84a62b73ab) to keep the code smaller and more consistent. This aligns with how the assets are treated in `board_plain.ts`.

If you prefer, I can extract the fix e4b6b1e13a4b7e3c4981bfd7522f150b0ea7fe41 for the url stones in the canvas renderer into another PR.

anoek:
Neat :) Could you attach an example custom board (or send them to anoek@online-go.com if you don't want them public), so I can test things out?

FYI there's another PR https://github.com/online-go/goban/pull/247 that I plan on merging soon, it might conflict with this one. Since it came first I'll merge it first, sorry in advance for the probable conflict, though I reckon Codex can get it sorted out.

lykahb:
Thank you for taking a look. I don't mind handling a merge conflict.

Here is a set of images:

Default
default-background

9x9
ogs-9x9

13x13
ogs-13x13

19x19
ogs-19x19

I had an alternative approach of how to handle multiple backgrounds that I decided against before submitting the PR, but I'd like to hear your opinion on it before merging.
- This PR hardcodes the assumption that the grid is at the certain coordinates of the image - the images with grid must have a margin of 1.5 squares. It is simple and gets the job done.
- Instead, we could describe the coordinates in the settings. However, that'd add complexity on the Themes&Visual page.

lykahb:
And for fun, here are the matching stones.

black-stone-16


white-stone-01

src/Goban/CanvasRenderer.ts
```diff
@@ -159,6 +163,19 @@ export class GobanCanvas extends Goban implements GobanCanvasInterface {
         /* TODO: Need to reconcile the clock fields before we can get rid of this `any` cast */
         super(config, preloaded_data as any);
 
+        this.grid_layer = createDeviceScaledCanvas(10, 10);
```
anoek: It can be problematic to allocate too many canvases, that's why we deal with conditional layers for things like shadows (and now the cross hairs from pr #247), let's apply that pattern for this case too so users who are not using this feature don't pay the extra cost.

anoek (changes requested):

src/Goban/CanvasRenderer.ts
lykahb: I've merged the upstream changes and added conditional layers, following the conventions of the crosshairs PR. Tested locally on both renderers.

anoek:
Thanks!

PR #257, opened 2026-07-01
Add custom board label color

lykahb:
Allow client to set the label color directly.

- Add a new callback `customBoardLabelColor`
- Move `blendWithInverseColor` to the `color.ts`. It is going to be reused by the label color reset on the settings page

Also see https://github.com/online-go/online-go.com/pull/3611


anoek:
Thanks!

PR #258, opened 2026-07-05
Add stone scale theme setting

lykahb:
This adds a new setting for the stone scaling, as I suggested at https://github.com/online-go/online-go.com/issues/3608

The canvas renderer limits the scaling factor to at most 1 to prevent overlapping.

See the matching PR https://github.com/online-go/online-go.com/pull/3612


anoek:
Thanks!

PR #260, opened 2026-07-15
Support custom stone image variants

lykahb:
This adds support for multiple custom black and white stone images in the SVG renderer, as I suggested at https://github.com/online-go/online-go.com/issues/3608. The Canvas renderer uses the first variant.

Stone variants are randomly distributed across intersections while remaining stable for the lifetime of each board. The variants may change on page reload. I think that the random arrangements work better, particularly for the themes where the variants differ more. Deterministic selection feels as if you always pick the same stone out of the bowl when playing the first moves. The default Shell theme still has the stable coordinate-based selection.

The change preserves compatibility with the existing single-URL callbacks and falls back to solid-color stones when custom images fail to load. Unit tests cover URL normalization, legacy compatibility, image failures, and stable variant placement.

Disclaimer: mostly written by Codex, extensively reviewed and edited in person

### DIFF
--- a/src/Goban/callbacks.ts
+++ b/src/Goban/callbacks.ts
@@ -47,7 +47,11 @@ export interface GobanCallbacks {
     customBoardLineColor?: () => string;
     customBoardLabelColor?: () => string;
     customBoardUrl?: () => string;
+    customBlackStoneUrls?: () => string[];
+    customWhiteStoneUrls?: () => string[];
+    /** @deprecated Use customBlackStoneUrls instead. */
     customBlackStoneUrl?: () => string;
+    /** @deprecated Use customWhiteStoneUrls instead. */
     customWhiteStoneUrl?: () => string;
 
     canvasAllocationErrorHandler?: (
@@ -70,8 +74,20 @@ export interface GobanCallbacks {
     toast?: (message_id: string, duration: number) => void;
 }
 
+function legacyBlackStoneUrls(): string[] {
+    const url = callbacks.customBlackStoneUrl?.().trim();
+    return url ? [url] : [];
+}
+
+function legacyWhiteStoneUrls(): string[] {
+    const url = callbacks.customWhiteStoneUrl?.().trim();
+    return url ? [url] : [];
+}
+
 export const callbacks: GobanCallbacks = {
     getClockDrift: () => 0,
+    customBlackStoneUrls: legacyBlackStoneUrls,
+    customWhiteStoneUrls: legacyWhiteStoneUrls,
 };
 
 /**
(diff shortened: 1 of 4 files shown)


lykahb:
Good point. I removed the image error handler and fallback. Custom SVG stones did not have special loading/error handling before this PR, and variants do not require it.

src/Goban/callbacks.ts
```diff
@@ -47,7 +47,13 @@ export interface GobanCallbacks {
     customBoardLineColor?: () => string;
     customBoardLabelColor?: () => string;
     customBoardUrl?: () => string;
+    /** Returns custom black stone image URLs, trimmed and unique within the array. */
+    customBlackStoneUrls?: () => string[];
+    /** Returns custom white stone image URLs, trimmed and unique within the array. */
+    customWhiteStoneUrls?: () => string[];
+    /** @deprecated Use customBlackStoneUrls instead. */
     customBlackStoneUrl?: () => string;
```
anoek: Should we go ahead and remove this now while we're here? We don't have any other consumers other than OGS using this feature so it should be safe to do. That'll clean out the "legacy" functions below too.

src/Goban/themes/image_stones.ts
```diff
@@ -48,6 +49,43 @@ type StoneType = {
...
+        !random_values ||
+        random_values.length !== goban.engine.height ||
+        random_values[0]?.length !== goban.engine.width
+    ) {
+        return initializeCustomStoneRandomValues(goban);
+    }
+    return random_values;
+}
+
+function getCustomStoneIndex(x: number, y: number, stones: unknown[], goban: GobanBase): number {
```
anoek: Traditionally we've computed our stone indicies using a combination of the game/review id and coordinates, that way the board looks the same when you load the same game. If I'm understanding this correctly, this is simply storing random values per gobanbase but that'll change every page load, so probably isn't what we want.

lykahb: The OGS shell theme looks good, but on my WIP themes the identical variants get easier to spot. Their stones are more distinct. I think that the identical variants shouldn't be adjacent often or at all. The Sabaki variant choice that I ported here looks noticeably better, but still can have adjacent variants.

I agree the changes to the variant choice goes beyond the scope of adding variants. I'm going to use the same coordinate-based method here.

Are you open to changes of the variant choice algorithm - as long as it is deterministic?

lykahb:
@anoek I've set the stone variant choice to reuse the existing coordinate-based algorithm

Here is a sneak peek of the latest theme in progress. 
image

